PR #27, opened 2021-11-11
RequestEvent: implement comment-type service

fenekku:
I am merging this directly like Max's previous PR, so that we have a common code base to work from from now on.

I've commented on numerous things that will need work.


invenio_requests/services/permissions.py
```diff
@@ -9,10 +9,74 @@
...
 """Request permissions."""
 
+from flask_principal import UserNeed
+from invenio_access.permissions import any_user
 from invenio_records_permissions import RecordPermissionPolicy
+from invenio_records_permissions.generators import Generator
 from invenio_records_permissions.generators import AnyUser, SystemProcess
 
 
+class Requesters(Generator):
```
fenekku, line 19: Once we define ownership, we will be able to flesh these out.

invenio_requests/services/permissions.py
```diff
@@ -21,3 +85,17 @@ class PermissionPolicy(RecordPermissionPolicy):
     can_create = [SystemProcess(), AnyUser()]
     can_update = [SystemProcess(), AnyUser()]
     can_delete = [SystemProcess(), AnyUser()]
+
+    # **Request Events Permission policy.**
+    # Comments need special cases for some actions
```
fenekku, line 90: I think that makes sense...

invenio_requests/services/schemas.py
```diff
@@ -0,0 +1,30 @@
...
+
+    type = fields.Str(
+        required=True,
+        validate=validate.OneOf(choices=[e.value for e in RequestEventType]),
+    )
+    content = utils_fields.SanitizedHTML()
+    format = fields.Str(
+        validate=validate.OneOf(choices=[e.value for e in RequestEventFormat]),
+        load_default=RequestEventFormat.HTML.value
+    )
```
fenekku, line 30: All RequestEvents have same shape for now. We'll see what happens when implementing the other kinds.

run-tests.sh
```diff
@@ -51,7 +51,8 @@ if [[ ${keep_services} -eq 0 ]]; then
 	trap cleanup EXIT
 fi
 
-black --check --diff invenio_requests tests
+# not related to the work at hand and different from other repos so commenting for now
+# black --check --diff invenio_requests tests
```
fenekku, line 55: .... I am sorry :bow: I put a higher value on running functionality test than style tests for now.

tests/conftest.py
```diff
@@ -81,13 +80,17 @@ def identity_simple():
...
 
 @pytest.fixture()
-def example_request(db, request_record_input_data):
+def example_request(db, identity_simple, request_record_input_data):
     """Example record."""
-    record = Request.create({}, **request_record_input_data)
-    record.commit()
-    db.session.commit()
-    return record
+    # Need to use the service to get the id I guess...
```
fenekku, line 92: yeah... is that normal?

tests/conftest.py
```diff
@@ -81,13 +80,17 @@ def identity_simple():
 @pytest.fixture()
 def request_record_input_data():
     """Input data to a Request record."""
-    return {}
+    return {
+        "title": "Foo bar",
+        "receiver": {"type": "baz", "id": "nyx"}
+    }
```
fenekku, line 86: Fakers will fake. Obviously replace this with something sensical.

tests/services/events/test_request_events_service.py
```diff
@@ -0,0 +1,72 @@
...
+    # Search (batch read) events
+    # first add another comment
+    data = copy.deepcopy(events_service_data)
+    data["content"] = "Another comment."
+    item2 = events_service.create(identity_simple, request_id, data)
+    # Refresh to make changes live
+    RequestEvent.index.refresh()
+    # then search
+    searched_items = events_service.search(
+        identity_simple, {"request_id": request_id}, size=10, page=1,
```
fenekku, line 64: This is very much on purpose as it decouples the REST API syntax from the service's call syntax.
The resource's responsibility will be to pass the right parameters to get the REST API output we want. The service doesn't care about those.

tests/services/events/test_request_events_service.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    assert 2 == searched_items.total
+
+
+def test_other_event_flow():
+    """Interact with other events."""
+    # events_service_data["type"] = RequestEventType.ACCEPT
+    pass
```
fenekku, line 72: TODO for other events.

fenekku (commented):
Comments map to next step work. Feel free to add more :)

PR #33, opened 2021-11-12
RequestEvent: implement REST API

### LINKED ISSUE #8: Events: initial version resource

fenekku:
closes #8

### DIFF
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021 TU Wien.
+# Copyright (C) 2021 Northwestern University.
 #
 # Invenio-Requests is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """API classes for requests in Invenio."""
 
-from collections import namedtuple
 from datetime import datetime
 from enum import Enum
 
@@ -175,10 +175,6 @@ class RequestEventFormat(Enum):
     HTML = "html"
 
 
-FakePID = namedtuple("FakePID", ["pid_value"])
-"""PID workaround."""
-
-
 class RequestEvent(Record):
     """A Request Event."""
 
@@ -190,6 +186,9 @@ class RequestEvent(Record):
     request = ModelField(dump=False)
     """The request."""
 
+    request_id = DictField("request_id")
+    """The data-layer id of the related Request."""
+
     type = ModelField("type")
     """The human-readable event type."""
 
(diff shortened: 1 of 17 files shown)


invenio_requests/records/api.py
```diff
@@ -151,6 +147,9 @@ class RequestEvent(Record):
     request = ModelField(dump=False)
     """The request."""
 
+    request_id = DictField("request_id")
```
fenekku, line 150: I didn't know how to make it work with RelatedModelField, so went for this which works.

invenio_requests/resources/events/resources.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    data_parser = request_body_parser(
+        parsers=from_conf('request_body_parsers'),
+        default_content_type=from_conf('default_content_type')
+    )
+
+    def create_url_rules(self):
+        """Create the URL rules for the record resource."""
+        # Aassignment of routes should be part of the
+        # Config class
```
fenekku, line 101: Just highlighting this. Would apply to the `from_conf` calls above too. This stuff should be taken care of for us by the config already at this point.

invenio_requests/services/events/config.py
```diff
@@ -10,15 +10,40 @@
...
+    def id(self):
+        """Id property."""
+        return self._record.id
+
+
+class RequestEventLink(Link):
+    """Link variables setter for RequestEvent links."""
+
+    # why static?
+    @staticmethod
```
fenekku, line 36: The one it overrides is static, don't know why...

invenio_requests/services/events/service.py
```diff
@@ -128,7 +136,7 @@ def delete(self, identity, id_, revision_id=None):
         self.require_permission(identity, permission, record=record)
 
         if record.type == RequestEventType.COMMENT.value:
-            record["type"] = RequestEventType.DELETED_COMMENT.value
+            record.type = RequestEventType.DELETED_COMMENT.value
```
fenekku, line 139: Leaky abstraction. We should go all in with field access I think more and more.

tests/resources/events/test_request_events_resources.py
```diff
@@ -0,0 +1,129 @@
...
+        json=events_resource_data
+    )
+
+    comment_id = response.json["id"]
+    expected_json_1 = {
+        **events_resource_data,
+        "id": comment_id,
+        "links": {
+            "self": f"https://127.0.0.1:5000/api/requests/{request_id}/comments/{comment_id}",  # noqa
+            # "self_html": "",  # TODO: UI link
```
fenekku, line 50: Should invidivual comments even have html links? Probably not.

PR #34, opened 2021-11-15
global: fix several issues

### LINKED ISSUE #13: Request API class: Rework (external) ID field
Currently, the system field for the external ID of the Request in the API class is called `id`, which is inconsistent with basically all other implementations of records: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/records/api.py#L34

This field (for the external ID) should probably be renamed into something else, maybe even `pid` (but without using an acutal `pidstore` model), to be more consistent with other record API classes?
And the basic `id` field should just point to the model's UUID.
Ping @lnielsen

### LINKED ISSUE #14: Requests: Implement created_by information
Currently, the creator information (`created_by`) is modeled in the JSON Schema and the ES mappings, but not touched in any other place.

We should create a system field for managing this information in the API class: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/records/api.py

This could potentially look similar to the `owners` part of the `access` system field (used in the parent records) from `rdm-records`: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/systemfields/access/owners.py
Maybe this could become a candidate for pulling up into a common dependency (e.g. `records-resources`).
*However*, I think that this field must be extendable – while the concept of `user` is defined way up in the dependency hierarchy, some concepts might be introduced only at a later (or even entirely unrelated) part of the hierarchy; for instance, `community`.
As such, we need some kind of extension mechanism for adding logic for resolving/adding further types.
Ping @lnielsen 

Also, we should set this information accordingly in the services, e.g. on create: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/services/requests/service.py#L66
Further, this information should likely be used in the permissions: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/services/permissions.py

**Additional information**

JSON schema: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/records/jsonschemas/requests/request-v1.0.0.json#L54-L67

Elasticsearch mappings: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/records/mappings/v7/requests/request-v1.0.0.json#L50-L57

### LINKED ISSUE #15: Request API class: Rework `request_type` field
Currently, the `request_type` constant field in the `Request` API class (https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/records/api.py#L41) is being used to choose the correct API class for handling any request that's loaded from storage in the service: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/services/requests/service.py#L41-L47

This is however not ideal for a few reasons:
* This field is intended to be human-readable, and not machine-readable (it's a string, which has potentially slow comparison operations)
* In no way is it even remotely guaranteed that multiple sub-classes of the basic `Request` API class have different `request_type`s set (the likelihood of a name collision in different Invenio extension modules is relatively high)
* It is somewhat likely that the value of these fields is changed at some point down the road (e.g. when a developer decides that they found a better name after a few releases), which is bad because this value is used for look-ups

Maybe some unique numeric value?
We could maybe provide a short documentation entry how to generate such an ID (e.g. hash from the package name + import path of the class), and stress that it's important to not change this field value down the road.

In the course of this work, we should also rework the way that custom request types are registered: https://github.com/inveniosoftware/invenio-requests/blob/de30a5f8e8cd25877718db2e776edc4f9025f740/invenio_requests/config.py#L26-L27

### LINKED ISSUE #20: Requests: Implement (and rename) `subject` field
Similar to https://github.com/inveniosoftware/invenio-requests/issues/14, the `subject` handling logic should be implemented.
Currently, it is only defined in the JSON schema and mappings(https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/jsonschemas/requests/request-v1.0.0.json#L33-L43), but it should also be used in the API class and the services.

As per discussion, we consider the current name (`subject`) to be overloaded and potentially confusing; it should be renamed to something like `associated_object`.

### LINKED ISSUE #21: Requests: Implement `receiver` field logic
Analogous to https://github.com/inveniosoftware/invenio-requests/issues/14 and https://github.com/inveniosoftware/invenio-requests/issues/20, the `receiver` field should be implemented.

### LINKED ISSUE #30: Schemas: Unify definition of "created_by", "receiver", and "subject"
To make things more consistent, we should unify the definition of `created_by`, `receiver` and `subject` because they do basically the same thing (reference some entity, where the set of referenceable entity types is not fixed).
That is, they should look like `{"user": id}` rather than `{"type": "user", "id": id}`.

In the JSON schema and mappings, we should probably just hint at the existence of *exactly one* property, and limit the possible shapes of the key (allow lowercase letters, maybe dashes, ...).

### LINKED ISSUE #32: Extensions: Move the request extension mechanisms
Currently, the extension mechanisms are placed in the API class (data layer), but the extension mechanisms (creating custom subtypes of requests with customized actions, etc.) does not entirely fit the data layer (custom request types should be easily writable for developers, and as such will probably use service functionality).
Thus, we should split the extension mechanisms (`available_actions`, etc.) into a separate class (`RequestType`?) and adjust the extension mechanism's workings accordingly.

This `RequestType` class should have a `.create(…)` method, similar to the `Record.create(…)`.

Also, a new "open" action should be made whose purpose it is to move the request out of its initial "draft" state.
For instance, in file replacement requests, this action would lock down the bucket and files submitted for replacement.

Note: The `Actions` should be atomic operations, like the service methods are (i.e. we don't want `db.commit()` in between operations, because if only later operations fail, we could end up in a committed inconsistent state).
Thus, some designing has to be done on how to solve this issue (maybe split up the service methods into smaller reusable chunks that do not call `db.commit()`?).

### LINKED ISSUE #37: Resource: Allow execution of request actions via REST API
The request actions should be executable via the REST API.

### LINKED ISSUE #38: Links: Generate links to the available actions for requests
The available action endpoints should be generated in the `links` section of the dumped request dictionary.
Bonus points if only the set of actions is generated that is actually available to the current user.

max-moser:
Addresses several issues.

closes #13 
closes #14 
closes #15
closes #20 
closes #21
closes #30 
closes #32
closes #37 
closes #38

### DIFF
--- a/invenio_requests/config.py
+++ b/invenio_requests/config.py
@@ -8,7 +8,7 @@
 
 """Invenio module for generic and customizable requests."""
 
-from .records.api import Request
+from .records.request_types import RequestType
 from .resolvers import UserResolver
 from .resolvers.requests import RequestResolver
 from .services.permissions import PermissionPolicy
@@ -16,9 +16,7 @@ from .services.permissions import PermissionPolicy
 REQUESTS_PERMISSION_POLICY = PermissionPolicy
 """Override the default requests/comments permission policy."""
 
-# TODO maybe only a list of classes?
-#      because we can get the 'request_type' strings from the classes
-REQUESTS_REGISTERED_TYPES = {Request.request_type.value: Request}
+REQUESTS_REGISTERED_TYPES = [RequestType()]
 """Configuration for registered Request Types."""
 
 REQUESTS_ENTITY_RESOLVERS = [
--- a/invenio_requests/config.py
+++ b/invenio_requests/config.py
@@ -9,6 +9,8 @@
 """Invenio module for generic and customizable requests."""
 
 from .records.api import Request
+from .resolvers import UserResolver
+from .resolvers.requests import RequestResolver
 from .services.permissions import PermissionPolicy
 
 REQUESTS_PERMISSION_POLICY = PermissionPolicy
@@ -18,3 +20,9 @@ REQUESTS_PERMISSION_POLICY = PermissionPolicy
 #      because we can get the 'request_type' strings from the classes
 REQUESTS_REGISTERED_TYPES = {Request.request_type.value: Request}
 """Configuration for registered Request Types."""
+
+REQUESTS_ENTITY_RESOLVERS = [
+    UserResolver(),
+    RequestResolver(),
+]
+"""Registered resolvers for resolving/creating references in request metadata."""
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -43,7 +43,7 @@ class Request(Record):
     )
     """Elasticsearch dumper with configured extensions."""
 
-    id = IdentityField("external_id")
+    number = IdentityField("external_id")
     """The request's external identity."""
 
     metadata = None
@@ -64,8 +64,8 @@ class Request(Record):
     custom request actions are registered.
     """
 
-    subject = ReferencedEntityField("subject")
-    """Subject (associated object) of the request."""
+    topic = ReferencedEntityField("topic")
+    """Topic (associated object) of the request."""
 
     created_by = ReferencedEntityField("created_by")
     """The entity that created the request."""
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -21,7 +21,12 @@ from .actions import AcceptAction, CancelAction, DeclineAction, ExpireAction
 from .dumpers import CalculatedFieldDumperExt
 from .models import RequestEventModel, RequestMetadata
 from .schema import RequestSchema
-from .systemfields import IdentityField, OpenStateCalculatedField, RequestStatusField
+from .systemfields import (
+    IdentityField,
+    OpenStateCalculatedField,
+    ReferencedEntityField,
+    RequestStatusField,
+)
 
 
 class Request(Record):
@@ -60,6 +65,15 @@ class Request(Record):
     To be overridden in subclasses, if the custom request type follows a
     different or more specific schema.
     """
+
+    subject = ReferencedEntityField("subject")
+    """Subject (associated object) of the request."""
+
+    created_by = ReferencedEntityField("created_by")
+    """The entity that created the request."""
+
+    receiver = ReferencedEntityField("receiver")
+    """The entity that will receive the request."""
 
     status = RequestStatusField("status")
     """The current status of the request."""
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -17,15 +17,14 @@ from invenio_records.systemfields import ConstantField, DictField, ModelField
 from invenio_records_resources.records.api import Record
 from invenio_records_resources.records.systemfields import IndexField
 
-from .actions import AcceptAction, CancelAction, DeclineAction, ExpireAction
-from .dumpers import CalculatedFieldDumperExt
+from .dumpers import CalculatedFieldDumperExt, RequestTypeDumperExt
 from .models import RequestEventModel, RequestMetadata
-from .schema import RequestSchema
 from .systemfields import (
     IdentityField,
     OpenStateCalculatedField,
     ReferencedEntityField,
     RequestStatusField,
+    RequestTypeField,
 )
 
 
@@ -33,37 +32,36 @@ class Request(Record):
     """A generic request record."""
 
     model_cls = RequestMetadata
+    """The model class for the request."""
 
     dumper = ElasticsearchDumper(
         extensions=[
             CalculatedFieldDumperExt("is_open"),
             CalculatedFieldDumperExt("is_expired"),
+            RequestTypeDumperExt("request_type"),
         ]
     )
+    """Elasticsearch dumper with configured extensions."""
 
     id = IdentityField("external_id")
+    """The request's external identity."""
 
     metadata = None
     """Disabled metadata field from the base class."""
 
-    # TODO figure out aliases, multiple mappings, common search fields
     index = IndexField("requests-request-v1.0.0", search_alias="requests")
+    """The Elasticsearch index to use for the request."""
 
     schema = ConstantField("$schema", "local://requests/request-v1.0.0.json")
+    """The JSON Schema to use for validation."""
 
-    request_type = ConstantField("request_type", "Generic Request")
-    """The human-readable request type.
+    request_type = RequestTypeField("request_type_id")
+    """System field for management of the request type.
 
-    To be overridden in subclasses with unique values, as this is used
-    in the RequestsService to choose the correct API class to use for a
-    request.
-    """
-
-    marshmallow_schema = RequestSchema
-    """Schema used for de/serialization of requests of this type.
-
-    To be overridden in subclasses, if the custom request type follows a
-    different or more specific schema.
+    This field manages loading of the correct RequestType classes associated with
+    `Requests`, based on their `request_type_id` field.
+    This is important because the `RequestType` classes are the place where the
+    custom request actions are registered.
     """
 
     subject = ReferencedEntityField("subject")
@@ -78,39 +76,11 @@ class Request(Record):
     status = RequestStatusField("status")
     """The current status of the request."""
 
-    available_statuses = {
-        "draft": True,
-        "open": True,
-        "cancelled": False,
-        "declined": False,
-        "accepted": False,
-        "expired": False,
-    }
-    """Available statuses for the Request.
-
-    The keys in this dictionary is the set of available statuses, and their
-    values are indicators whether this Request is still considered to be
-    "open" in this state.
-    """
-
-    available_actions = {
-        "accept": AcceptAction,
-        "cancel": CancelAction,
-        "decline": DeclineAction,
-        "expire": ExpireAction,
-    }
-    """Available actions for this Request.
-
-    The keys are the internal identifiers for the actions, the values are
-    the actual RequestAction classes (not objects).
-    Whenever an action is looked up, a new object of the registered
-    RequestAction class is instantiated with the current Request object as
-    argument.
-    """
-
     is_open = OpenStateCalculatedField("is_open")
+    """Whether or not the current status can be seen as an 'open' state."""
 
     expires_at = ModelField("expires_at")
+    """Expiration date of the request."""
 
     @property
     def is_expired(self):
@@ -128,12 +98,29 @@ class Request(Record):
         return d < now
 
     def get_action(self, action_name):
-        return self.available_actions[action_name](self)
+        """Get the action registered under the given name.
+
+        :param action_name: The registered name of the action.
+        :return: The action registered under the given name.
+        """
+        return self.request_type.available_actions[action_name](self)
 
     def can_execute_action(self, action_name, identity):
+        """Check if the action registered under the given name can be executed.
+
+        :param action_name: The registered name of the action.
+        :param identity: The identity who wants to execute the action.
+        :return: Whether or not the action can be executed.
+        """
         return self.get_action(action_name).can_execute(identity)
 
     def execute_action(self, action_name, identity):
+        """Execute the action registered under the given name.
+
+        :param action_name: The registered name of the action.
+        :param identity: The identity who wants to execute the action.
+        :return: The return value of the executed action.
+        """
         return self.get_action(action_name).execute(identity)
 
 
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -17,6 +17,7 @@ from invenio_records.systemfields import ConstantField, DictField, ModelField
 from invenio_records_resources.records.api import Record
 from invenio_records_resources.records.systemfields import IndexField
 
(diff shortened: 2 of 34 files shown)


invenio_requests/records/schema.py
```diff
@@ -21,14 +21,10 @@ class AgentSchema(Schema):
...
-
-class ObjectSchema(Schema):
-    """Schema for a generic object."""
-
-    type = SanitizedUnicode(required=True)
-    id = SanitizedUnicode(required=True)
+    # TODO will need to accomodate for communities (as receivers), records (as
+    #      subjects/associated objects) and probably a few more
+    #      - but only for one of them at a time!
+    user = fields.String(required=True)
```
max-moser, line 27: the agent schema should be extensible, maybe via some kind of registry?
e.g. `invenio-communities` should be able to extend it by allowing a `community` key, with a `comid` (string?) as value.

note that those properties should be mutually exclusive (i.e. only one property should be defined at a time)!

invenio_requests/records/mappings/v6/requests/request-v1.0.0.json
```diff
@@ -2,6 +2,32 @@
...
+        },
+        {
+          "receivers": {
+            "path_match": "receiver.*",
+            "mapping": {
+              "type": "keyword"
+            }
+          }
+        }
+      ],
```
max-moser, line 30: restrict dynamically added properties for references to be always of type `keyword`, via dynamic templates: https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html#path-match-unmatch
(maybe this could be collapsed into a single dynamic template?)
this should allow us to reference all kinds of entities

note: although the overall `dynamic` setting is set to `strict`, properties of these fields do allow dynamic values (see below)

invenio_requests/records/jsonschemas/requests/definitions-v1.0.0.json
```diff
@@ -0,0 +1,22 @@
...
+  "agent": {
+    "description": "Reference to an entity, with the type as key and ID as value",
+    "type": "object",
+    "additionalProperties": false,
+    "patternProperties": {
+      "^[a-z]+$": {
+        "type": "string"
+      }
+    }
+  },
```
max-moser, line 11: very generic schema for allowing references to not only `user`s, but also currently still unknown entity types
since we need to allow references to non-acting entities as well (e.g. records, in the `subject` field [which is to be renamed to `associated_object`]), we should maybe consider renaming this definition?
`reference`? `entity-reference`?

invenio_requests/records/request_types.py
```diff
@@ -0,0 +1,57 @@
...
+    }
+    """Available statuses for the Request.
+
+    The keys in this dictionary is the set of available statuses, and their
+    values are indicators whether this Request is still considered to be
+    "open" in this state.
+    """
+
+    available_actions = {
+        "accept": AcceptAction,
```
max-moser, line 32: we still need another action for getting the request out of `draft` state (`SubmitAction`?)

invenio_requests/records/schema.py
```diff
@@ -40,8 +36,8 @@ class RequestSchema(BaseRecordSchema):
     payload = fields.Dict(dump_only=True)
 
     created_by = fields.Nested(AgentSchema, dump_only=True)
-    receiver = fields.Nested(ObjectSchema, required=True)
-    subject = fields.Nested(ObjectSchema)
+    receiver = fields.Nested(AgentSchema, required=True)
```
max-moser, line 39: in the resolvers, i started referring to these kinds of dicts as "reference dicts"

invenio_requests/resolvers/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+"""Module for resolvers."""
+
+from flask import current_app
+
+from .base import EntityResolver
+from .helpers import reference_entity, resolve_entity
+from .records import RecordResolver
+
+# from .requests import RequestResolver
```
max-moser, line 17: would create a circular import with the entity resolver system field; probably the structure should be refactored here

invenio_requests/resolvers/records.py
```diff
@@ -0,0 +1,21 @@
...
+# details.
+
+from .base import EntityResolver
+
+
+class RecordResolver(EntityResolver):
+    ENTITY_TYPE_KEY = "record"
+    ENTITY_TYPE_CLASS = None
+
+    def __init__(self, record_cls):
```
max-moser, line 16: can be instantiated with `RecordResolver(record_cls=RDMRecord)`, then it should be good to go i think (needs testing though)

lnielsen (approved):

invenio_requests/config.py
```diff
@@ -8,13 +8,19 @@
...
+from .resolvers.requests import RequestResolver
 from .services.permissions import PermissionPolicy
 
 REQUESTS_PERMISSION_POLICY = PermissionPolicy
 """Override the default requests/comments permission policy."""
 
-# TODO maybe only a list of classes?
-#      because we can get the 'request_type' strings from the classes
-REQUESTS_REGISTERED_TYPES = {Request.request_type.value: Request}
+REQUESTS_REGISTERED_TYPES = [RequestType()]
```
max-moser, line 19: this could probably be reworked to use `entrypoints` for registration...
i think there's no real harm in having a couple unused RequestTypes defined by installed packages
the request type registry will turn this into a dictionary for lookups anyway

invenio_requests/resolvers/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+        if resolver.matches_entity(entity):
+            return resolver.reference(entity, check=False)
+
+    if raise_:
+        raise ValueError(f"no matching resolver registered for: {type(entity)}")
+
+    return None
+
+
+def reference_identity(identity, raise_=False):
```
max-moser, line 49: this still needs a bit of love, as it's currently either a `user` or the `system_identity` – for which we don't dump anything at all, similar to the `AccessComponent` in RDM-Records: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/components/access.py#L35-L45

i think the resolvers could do something similar with identities as they're currently doing with entity classes (records, users, ...)

invenio_requests/records/schema.py
fenekku: This sounds like a https://github.com/marshmallow-code/marshmallow-oneofschema perhaps?

invenio_requests/services/requests/links.py
```diff
@@ -17,5 +17,5 @@ class RequestLink(Link):
     @staticmethod
     def vars(record, vars):
         """Variables for the URI template."""
-        # TODO
-        vars.update({"id": record.id or record.model.id})
+        # TODO this uses the UUID of the record, should we maybe use the number/ext-id?
```
fenekku, line 20: Yes, we should use the number/external-id because this is used in the links which are user facing and supposed to be persistent.

invenio_requests/services/requests/results.py
```diff
@@ -32,11 +32,14 @@ def __init__(
...
-        self._schema = schema or service._wrap_schema(request.marshmallow_schema)
+        self._schema = schema or service._wrap_schema(
+            request.request_type.marshmallow_schema
+        )
 
     @property
     def id(self):
-        return self._request.id
+        """Identity of the request."""
+        return self._request.number
```
fenekku, line 42: :heavy_check_mark:

invenio_requests/services/requests/service.py
```diff
@@ -63,22 +44,27 @@ def _get_request(self, id_):
...
-        schema = self._wrap_schema(request_class.marshmallow_schema)
+        schema = self._wrap_schema(request_type.marshmallow_schema)
         data, errors = schema.load(
             data,
             context={"identity": identity},
         )
 
-        # it's the components that will populate the actual data
-        request = request_class.create({})
+        # it's the components that will populate most of the actual data
```
fenekku, line 59: is that still true? ;)

invenio_requests/records/api.py
```diff
@@ -114,12 +98,29 @@ def is_expired(self):
         return d < now
 
     def get_action(self, action_name):
-        return self.available_actions[action_name](self)
+        """Get the action registered under the given name.
+
+        :param action_name: The registered name of the action.
+        :return: The action registered under the given name.
+        """
+        return self.request_type.available_actions[action_name](self)
```
fenekku, line 106: Returning a specialized error that can be caught at Resource level might be helpful here to return a 404 when the action is unknown.

tests/conftest.py
```diff
@@ -68,6 +69,21 @@ def create_app(instance_path):
...
+    with db.session.begin_nested():
+        datastore = app.extensions["security"].datastore
+        user = datastore.create_user(
+            email="user@inveniosoftware.org",
+            password=hash_password("password"),
+            active=True,
+        )
+
+    db.session.commit()
+    return user
```
fenekku, line 84: See what I did here: https://github.com/inveniosoftware/invenio-requests/pull/33/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R110 . I can change yours for that if you think it's helpful.

fenekku (approved):
I can't comment too much on the RequestType - action logic (opaque to me), but it looks good. I will let you take a look and tomorrow we can decide what to merge.

invenio_requests/records/jsonschemas/requests/definitions-v1.0.0.json
```diff
@@ -0,0 +1,22 @@
+{
+  "entity_reference": {
+    "description": "Reference to an entity, with the type as key and ID as value",
+    "type": "object",
+    "additionalProperties": false,
+    "patternProperties": {
+      "^[a-z]+$": {
```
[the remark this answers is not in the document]
max-moser: fixed!

invenio_requests/services/requests/service.py
max-moser: at least the external identity is set via a component :stuck_out_tongue: 
i've updated the comment accordingly!

tests/conftest.py
max-moser: yes, multiple users will come in handy at the latest when we're adding tests for permissions (e.g. one user sends a request to another user, and we test that each user only has permissions on a subset of the actions)

invenio_requests/services/requests/links.py
max-moser: IIRC, the ID to use would depend on the endpoint in question (depends on where we can use the `number` nicely)?

since the `number` construction logic is customizable, we can create it like `f"{comid}:{number-suffix}"` and thus use it in endpoints like `/api/communities/<comid>/requests/<number-suffix>`
in the "general" `/api/requests/<id>` endpoint, the `id` in use would be the internal UUID?

ping @lnielsen

invenio_requests/records/schema.py
max-moser: thanks for the pointer!
i'll create an issue for investigating this further

invenio_requests/records/jsonschemas/requests/definitions-v1.0.0.json
```diff
@@ -0,0 +1,22 @@
...
+        "type": "string"
+      }
+    }
+  },
+  "anything": {
+    "description": "A very general schema",
```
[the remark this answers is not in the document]
max-moser: i was thinking that we could maybe enforce some kind of prefix for the items in there, to avoid type clashes for ES index mappings
however, i haven't spent much time on the `payload` yet, as that's still unused... i think this would need some scrutiny in the future anyway

invenio_requests/records/api.py
```diff
@@ -17,86 +17,70 @@
...
-    request.
+    This field manages loading of the correct RequestType classes associated with
+    `Requests`, based on their `request_type_id` field.
+    This is important because the `RequestType` classes are the place where the
+    custom request actions are registered.
     """
 
-    marshmallow_schema = RequestSchema
-    """Schema used for de/serialization of requests of this type.
+    topic = ReferencedEntityField("topic")
```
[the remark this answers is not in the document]
max-moser: you mean, we could ditch the string in the constructor because we could get this reliably from somewhere else?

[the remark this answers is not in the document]
fenekku: Yes, that string corresponds to the attribute name in each case and this is what Python data descriptors are about (why else were we using the data descriptors for :sweat_smile: ?)

PR #35, opened 2021-11-15
requests: move read functionality from service to api class

### LINKED ISSUE #25: Requests: Move logic for resolving Request object by (external) ID to the API class
Currently, the logic for resolving a request by ID lives in the `RequestsService`: https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/services/requests/service.py#L49

However, this necessitates that whoever wants to resolve `Request`s has to have a reference to some `RequestsService`.
To remove this requirement, we should probably move this logic to the API class, probably by overriding the already existing `.get_record(…)` method: https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L370

max-moser:
Because it was awkward to require a RequestsService to fetch a Request object from database via its ID

closes #25

### DIFF
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -12,10 +12,12 @@ from datetime import datetime
 from enum import Enum
 
 import pytz
+from invenio_db import db
 from invenio_records.dumpers import ElasticsearchDumper
 from invenio_records.systemfields import ConstantField, DictField, ModelField
 from invenio_records_resources.records.api import Record
 from invenio_records_resources.records.systemfields import IndexField
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
 from ..errors import NoSuchActionError
 from .dumpers import CalculatedFieldDumperExt, RequestTypeDumperExt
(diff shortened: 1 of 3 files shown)


invenio_requests/records/api.py
```diff
@@ -122,6 +125,48 @@ def can_execute_action(self, action_name, identity):
...
+
+    @classmethod
+    def get_record(cls, id_, with_deleted=False):
+        """Retrieve the request by id.
+
+        :param id_: The record ID (external or internal).
+        :param with_deleted: If `True`, then it includes deleted requests.
+        :returns: The :class:`Request` instance.
+        """
+        with db.session.no_autoflush:
```
fenekku, line 150: Why `no_autoflush` here? I understand its use when creating model instances, but when doing a query?

invenio_requests/records/api.py
```diff
@@ -122,6 +125,48 @@ def can_execute_action(self, action_name, identity):
...
+            try:
+                query = cls.model_cls.query.filter_by(external_id=str(id_))
+                if not with_deleted:
+                    query = query.filter(cls.model_cls.is_deleted != True)
+
+                model = query.one()
+
+            except (MultipleResultsFound, NoResultFound):
+                # either no results or ambiguous results
+                # (e.g. if external_id is None)
```
fenekku, line 160: How could the external_id be None? Isn't it the id used in the URL of the request?

invenio_requests/services/events/service.py
```diff
@@ -196,10 +196,7 @@ def _get_permission(self, action, event_type):
 
     def _get_request(self, request_id):
         """Get associated request."""
-        # TODO: Replace with whatever interface Request provides to get one
-        from ..requests import RequestsService, RequestsServiceConfig
-
-        return RequestsService(config=RequestsServiceConfig())._get_request(request_id)
+        return self.request_cls.get_record(request_id)
```
fenekku, line 199: :+1:

fenekku (commented):
Just some clarifications on this one. Thanks for the getter!

invenio_requests/records/api.py
max-moser: i basically just copied over the `get_record` method from `invenio-records` and adapted it: https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L379
the exact reasoning is not clear to me either, but since it's done that way in the base implementation, i thought it'd be better to keep it.

invenio_requests/records/api.py
max-moser: if the requests are created via the `requests_service.create(…)`, the `external_id` should never be `None` because the `ExternalIdentifierComponent` takes care of initializing that.
In order for it to be `None`, it would have to have been (incompletely) created through other means as the service, **or** if the custom logic for creating the external identifier (aka number) would somehow return a `None` value.

invenio_requests/records/api.py
```diff
@@ -122,6 +125,48 @@ def can_execute_action(self, action_name, identity):
     def execute_action(self, action_name, identity):
         return self.get_action(action_name).execute(identity)
 
+    @classmethod
+    def _request_from_model(cls, model):
+        """Create a Request object from the given model."""
+        # TODO handle 'model is None' more gracefully
+        #      -> may happen e.g. when reading a deleted request
```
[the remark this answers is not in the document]
max-moser: maybe, yes
could that be some kind of a tombstone analogue for requests?

PR #44, opened 2021-11-18
Request: comment on submit and accept

fenekku:
- closes https://github.com/inveniosoftware/invenio-requests/issues/41


invenio_requests/actions.py
```diff
@@ -0,0 +1,196 @@
...
+        self.request.status = "accepted"
+
+        request_id = self.request.number
+
+        # persist changes
+        self.request.commit()
+        db.session.commit()
+        requests_service = current_requests.requests_service
+        if requests_service.indexer:
+            requests_service.indexer.index(self.request)
```
fenekku, line 137: this part feels like it will be repeated...

invenio_requests/config.py
```diff
@@ -16,7 +23,15 @@
...
-REQUESTS_REGISTERED_TYPES = [RequestType()]
+available_actions = {
+    "submit": SubmitAction,
+    "accept": AcceptAction,
+    "cancel": CancelAction,
+    "decline": DeclineAction,
+    "expire": ExpireAction,
+}
+
+REQUESTS_REGISTERED_TYPES = [RequestType(available_actions)]
```
fenekku, line 34: Addressed the circular import caused by extracting `actions.py` outside of `records/`, by injecting the actions at initialization time... maybe there is a better workflow.

invenio_requests/services/events/service.py
```diff
@@ -191,10 +193,16 @@ def request_cls(self):
...
+        Needed to distinguish between kinds of events.
         """
         if event_type == RequestEventType.COMMENT.value:
             return f"{action}_event_comment"
+        elif event_type == RequestEventType.ACCEPTED.value:
+            return f"accept"
+        elif event_type == RequestEventType.DECLINED.value:
+            return f"decline"
+        elif event_type == RequestEventType.CANCELLED.value:
+            return f"cancel"
```
fenekku, line 205: We will need something more flexible to support different available actions (different request types). Convenient for now.

tests/conftest.py
```diff
@@ -84,7 +83,7 @@ def identity_simple():
 @pytest.fixture(scope="module")
 def request_record_input_data():
     """Input data to a Request record."""
-    return {"title": "Foo bar", "receiver": {"user": "1"}}
+    return {"title": "Foo bar", "receiver": {"user": "2"}}
```
fenekku, line 86: just setting up the receiver to be different than the typical sender

tests/services/requests/test_requests_service.py
```diff
@@ -0,0 +1,128 @@
...
+        "content": "Welcome to the community!",
+        "format": RequestEventFormat.HTML.value
+    }
+    result = requests_service.execute_action(identity_simple_2, id_, "accept", data)
+    result_dict = result.to_dict()
+
+    RequestEvent.index.refresh()
+
+    assert "accepted" == result_dict["status"]
+    results = request_events_service.search(identity_simple_2, id_)
```
fenekku, line 117: Note that ordering will be dealt with at the resource-level.

fenekku (commented):
Notes for next PRs.

invenio_requests/config.py
max-moser: i've managed to get rid of this circular import by removing the `import RequestType` in the `RequestTypeField`: https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/systemfields/request_type.py#L13

probably splitting up our current "default implementation" of `RequestType` (https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/request_types.py#L28) into a semantically empty base class and a sub-class thereof with our current implementation (into separate files) could do the trick
then we could still use our empty base class (which wouldn't cause circular imports because it doesn't need to import anything) in the system field for type checks in the system field
and we could still use the default `RequestType` implementation in the config

max-moser (approved):

PR #45, opened 2021-11-19
Refactor and clean up

max-moser:
* move the parts that are intended for customization to their own modules
* split the base classes and the default implementations (which also fixes the circular import)
* fix issues detected by flake8 (e.g. unused or missing imports)
* format the code with black and isort

The most important part in this PR is that the parts of the request customization mechanism (i.e. `RequestType`, `RequestAction`, `RequestSchema`) have been moved to their separate module (`customization`) and split up into base classes and default implementations (`customization.{base,default}`).


invenio_requests/customizations/base/actions.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# Invenio-Requests is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Base class for customizable actions on requests."""
+
+from ...errors import CannotExecuteActionError, NoSuchActionError
+
+
+class RequestAction:
```
fenekku, line 14: I am relatively new to this, but especially for customization classes that would be written by other developers, [Python Abstract Base Classes](https://docs.python.org/3/library/abc.html) are a really nice touch. They will alert the customizers before code is run that their class definitions (well instances) are not implementing all the required interfaces. So e.g.,

```python
import abc

class RequestAction(abc.ABC):
    def __init__(self, request):
        self.request = request

    # have to override
    @abc.abstractmethod
    def can_execute(self, identity, data=None):
        """..."""

    # have to override
    @abc.abstractmethod
    def execute(self, identity, data=None):
        """..."""

    # don't have to override
    def post_execute(self, identity, data=None):
        """..."""
```

Would create this experience for the customizer:

```python
# customizer_code.py

class InvalidRequestActionImplementation(RequestAction):
    def execute(self, identity, data=None):
        """Execute!"""
        print("Implemented execute")

class ValidRequestActionImplementation(RequestAction):
    def can_execute(self, identity, data=None):
        """Can Execute?"""
        print("Implemented can_execute")
        return True

    def execute(self, identity, data=None):
        """Execute!"""
        print("Implemented execute")
          
a = InvalidRequestActionImplementation("request")
# Traceback (most recent call last):
#   File "test_abc.py", line 39, in <module>
#     a = InvalidRequestActionImplementation()
# TypeError: Can't instantiate abstract class InvalidRequestActionImplementation with abstract methods can_execute

b = ValidRequestActionImplementation("request")
```

fenekku (approved):
:+1: I leave it open for you to consider adding ABC. Merge right away if you want to add this later.

invenio_requests/customizations/base/actions.py
max-moser: i'm all for this!

fenekku:
Another comment (delete me!)

PR #49, opened 2021-11-19
40 timeline link

### LINKED ISSUE #40: Add timeline link to timeline response json
i.e. `"self": "https://127.0.0.1:5000/api/requests/<request_id>/timeline?page=1&size=25"

fenekku:
(
- Request: comment on cancel and decline
- RequestEvent: rework delete semantics
- RequestEvent: guarantee type immutability
)
- RequestEvent: dump search links; closes #40

### DIFF
--- a/invenio_requests/services/events/config.py
+++ b/invenio_requests/services/events/config.py
@@ -8,10 +8,12 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Request Events Service Config."""
+from functools import partial
 
 from invenio_records_resources.services import RecordServiceConfig
 from invenio_records_resources.services.base.links import Link
 from invenio_records_resources.services.records.components import DataComponent
+from invenio_records_resources.services.records.links import pagination_links
 from invenio_records_resources.services.records.results import RecordItem
 
 from ...records.api import Request, RequestEvent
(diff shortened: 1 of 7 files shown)


invenio_requests/services/events/config.py
```diff
@@ -58,3 +60,4 @@ class RequestEventsServiceConfig(
     links_item = {
         "self": RequestEventLink("{+api}/requests/{request_id}/comments/{id}"),
     }
+    links_search = pagination_links("{+api}/requests/{request_id}/timeline{?args*}")
```
fenekku, line 63: :seedling:  `pagination_links` creates `Link` objects out of the template. But since the LinksTemplate is generated by the service(s) with the desired context, overriding Link to provide custom `Links.vars()` is probably not necessary. But that static method is enticing by its mere existence. I think we can get rid of it in invenio-records-resources and use the context all the time...

invenio_requests/services/events/service.py
```diff
@@ -119,10 +119,10 @@ def update(self, identity, id_, data, revision_id=None):
     def delete(self, identity, id_, revision_id=None):
         """Delete an event from database and search indexes.
 
-        Deleting a comment is wiping the content and setting the type
-        to deleted.
+        Deleting a comment is wiping the content and marking it deleted.
+        Deleting an accepted/cancelled/declined is just wiping the content.
+        Deleting another event is really deleting them.
```
fenekku, line 124: To be honest, the semantics of delete can be changed to whatever we want as long as there is clarity on them.

fenekku:
closed in favor of https://github.com/inveniosoftware/invenio-requests/pull/53

PR #50, opened 2021-11-22
Just review comments for discussion (don't merge)

lnielsen:


invenio_requests/records/api.py
```diff
@@ -20,6 +20,7 @@
 
 from .dumpers import CalculatedFieldDumperExt, RequestTypeDumperExt
 from .models import RequestEventModel, RequestMetadata
+# import style
```
max-moser, line 23: will have to look into how to change that in black & isort
if the `import` lines get too long, they're styled the same way in `flask-resources` too, though: https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/__init__.py#L151-L157

PR #53, opened 2021-11-22
RequestEvent: dump search links

fenekku:
closes: https://github.com/inveniosoftware/invenio-requests/issues/40

### DIFF
--- a/invenio_requests/services/events/config.py
+++ b/invenio_requests/services/events/config.py
@@ -8,10 +8,12 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Request Events Service Config."""
+from functools import partial
 
 from invenio_records_resources.services import RecordServiceConfig
 from invenio_records_resources.services.base.links import Link
 from invenio_records_resources.services.records.components import DataComponent
+from invenio_records_resources.services.records.links import pagination_links
 from invenio_records_resources.services.records.results import RecordItem
 
 from ...records.api import Request, RequestEvent
@@ -54,3 +56,4 @@ class RequestEventsServiceConfig(RecordServiceConfig, CustomizationConfigMixin):
     links_item = {
         "self": RequestEventLink("{+api}/requests/{request_id}/comments/{id}"),
     }
+    links_search = pagination_links("{+api}/requests/{request_id}/timeline{?args*}")
--- a/tests/resources/events/test_request_events_resources.py
+++ b/tests/resources/events/test_request_events_resources.py
@@ -15,8 +15,8 @@ from invenio_requests.records.api import RequestEvent, RequestEventType
 
 
 def assert_api_response_json(expected_json, received_json):
-    """Assert the REST API response"s json."""
-    # We don"t compare dynamic times at this point
+    """Assert the REST API response's json."""
+    # We don't compare dynamic times at this point
     received_json.pop("created")
     received_json.pop("updated")
     assert expected_json == received_json
--- a/tests/resources/events/test_request_events_resources.py
+++ b/tests/resources/events/test_request_events_resources.py
@@ -15,8 +15,8 @@ from invenio_requests.records.api import RequestEvent, RequestEventType
 
 
 def assert_api_response_json(expected_json, received_json):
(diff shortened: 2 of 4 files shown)


invenio_requests/services/events/config.py
```diff
@@ -54,3 +56,4 @@ class RequestEventsServiceConfig(RecordServiceConfig, CustomizationConfigMixin):
     links_item = {
         "self": RequestEventLink("{+api}/requests/{request_id}/comments/{id}"),
     }
+    links_search = pagination_links("{+api}/requests/{request_id}/timeline{?args*}")
```
fenekku, line 59: :seedling:  `pagination_links` creates `Link` objects out of the template. But since the LinksTemplate is generated by the service(s) with the desired context, overriding Link to provide custom `Links.vars()` is probably not necessary. But that static method is enticing by its mere existence. I think we can get rid of it in invenio-records-resources and use the context all the time...

tests/resources/events/test_request_events_resources.py
```diff
@@ -15,8 +15,8 @@
 
 
 def assert_api_response_json(expected_json, received_json):
-    """Assert the REST API response's json."""
-    # We don't compare dynamic times at this point
+    """Assert the REST API response"s json."""
+    # We don"t compare dynamic times at this point
```
max-moser, line 19: :eyes:

invenio_requests/services/events/config.py
max-moser: ping @lnielsen

max-moser (approved):

invenio_requests/services/events/config.py
lnielsen: I'm not sure I understand 🤔The purpose of ``pagination_links`` is a shortcut for writing something like below:

```python
links_search = {
     "self": Link("{+api}/requests/{request_id}/timeline{?args*}", when=....),
     "prev": Link("{+api}/requests/{request_id}/timeline{?args*}", when=....),
     "next": Link("{+api}/requests/{request_id}/timeline{?args*}", when=....),
}
```

tests/resources/events/test_request_events_resources.py
```diff
@@ -15,8 +15,8 @@
 
 
 def assert_api_response_json(expected_json, received_json):
-    """Assert the REST API response's json."""
-    # We don't compare dynamic times at this point
+    """Assert the REST API response"s json."""
```
lnielsen, line 18:
```suggestion
    """Assert the REST API response's json."""
```

tests/resources/events/test_request_events_resources.py
lnielsen:
```suggestion
    # We don't compare dynamic times at this point
```

lnielsen (approved):

invenio_requests/services/events/config.py
fenekku: @lnielsen ha yes re-reading that wasn't super clear :sweat_smile: . What I meant was that `pagination_links` does exactly what you outlined: it creates a **Link** object for "self", "prev" and "next" but a `Link` object is sometimes not appropriate since a `RequestLink` or some such variant can be defined to fill in the `{request_id}` or additional parameters. These variants are billed as shortcuts (shortcut to `LinksTemplate(self.config.links_item, context=...)`) and encouraged by 
```python
    @staticmethod
    def vars(obj, vars):
        """Subclasses can overwrite this method."""
        pass
``` 
in `Link`'s definition. But pagination_links doesn't support them.

In response to this situation that is confusing (for item routes you can or not, create your own `LinkChild`, but for list routes you can't unless you create your own pagination_links  ... ) or sometimes just wrong (the link template won't expand correctly if the right Link isn't chosen), **my recommendation would be to remove that `vars` staticmethod** above altogether. This way there is less decision to be made by the developer and it's clearer: always use `LinksTemplate(self.config.links_item, context=...)` in the service. Plus there is already a "brevity" method `self.links_item_tpl` in the service that can be overridden to save some repetition.

PR #55, opened 2021-11-22
RequestEvent: define full data model (REST to ES)

### LINKED ISSUE #54: Define shape of RequestEvent in API, data-layer, DB and ES
From discussion, the RequestEvents will have this shape:

```python
{
    "id": ...,
    "revision_id": 2,
    "created": ...,
    "modified": ...,
    "request_id": ...,
    "created_by": ...,
    
    "type": "comment",
    "extra": {
        "format": "html",    
        "content": "bla bla bla",
    }
}    
```

with `extra` varying per `type`

fenekku:
closes #54


invenio_requests/records/jsonschemas/request_events/event-v1.0.0.json
```diff
@@ -0,0 +1,24 @@
...
+    "id": {
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
+    },
+    "request_id": {
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
+    },
+    "type": {
+      "type": "string"
+    },
+    "extras": {
```
lnielsen, line 19: Can we name it ``payload`` to align with the [request schema](https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/jsonschemas/requests/request-v1.0.0.json#L25)

invenio_requests/records/jsonschemas/request_events/event-v1.0.0.json
```diff
@@ -0,0 +1,24 @@
...
+      "type": "object",
```
lnielsen, line 20: This puts a bit harder structure on the keys.
```suggestion
      "$ref": "local://requests/definitions-v1.0.0.json#/anything"
```

invenio_requests/records/mappings/v6/request_events/event-v1.0.0.json
```diff
@@ -0,0 +1,41 @@
...
+        },
+        "request": {
+          "type": "keyword"
+        },
+        "request_id": {
+          "type": "keyword"
+        },
+        "extras": {
+          "type": "object",
+          "dynamic": true
```
lnielsen, line 36: I think we should skip indexing of the extras/payload: 
```suggestion
          "enabled": false
```

invenio_requests/services/schemas.py
```diff
@@ -24,15 +25,45 @@ def get_value(self, obj, attr, **kwargs):
         return getattr(obj, attr, obj.get(attr, missing))
 
 
-class RequestEventSchema(BaseRecordSchema):
+class BaseEventSchema(BaseRecordSchema):
```
lnielsen, line 28: Comment: We should align on where to put the schemas - I prefer here, but we also have https://github.com/inveniosoftware/invenio-requests/blob/528865dd5db0bd66e8ea8e876e42b445a56063f2/invenio_requests/customizations/base/schema.py

lnielsen (approved):

invenio_requests/records/mappings/v6/request_events/event-v1.0.0.json
fenekku: We won't be able to perform search on those which I think is fine for now.

invenio_requests/services/schemas.py
fenekku: I do too as per how we've done it in other invenio-modules e.g., https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/schema.py or https://github.com/inveniosoftware/invenio-rdm-records/tree/master/invenio_rdm_records/services/schemas . I will move https://github.com/inveniosoftware/invenio-requests/blob/528865dd5db0bd66e8ea8e876e42b445a56063f2/invenio_requests/customizations/base/schema.py in here now as part of this PR.

fenekku:
In test_alembic.py, `assert len(ext.alembic.compare_metadata()) == 0` fails locally but passes on CI. Something isn't right here.

PR #56, opened 2021-11-23
actions: support unit of work and minor refactor

lnielsen:


invenio_requests/customizations/base/actions.py
```diff
@@ -29,35 +29,16 @@ def can_execute(self, identity, data=None):
...
 
         :param identity: The identity of the executor.
         :param data: The passed input to the action.
         """
-        # probably want to do something with self.request.topic
-        if not self.can_execute(identity):
-            action_name = type(self).__name__
-            raise CannotExecuteActionError(action_name)
-
-    def post_execute(self, identity):
```
lnielsen, line 45: Comment: With ``uow`` we won't need this anymore.

invenio_requests/customizations/base/actions.py
```diff
@@ -29,35 +29,16 @@ def can_execute(self, identity, data=None):
...
 
-    def execute(self, identity, data=None):
+    def execute(self, identity, data=None, uow=None):
         """Execute the request action.
 
         :param identity: The identity of the executor.
         :param data: The passed input to the action.
         """
-        # probably want to do something with self.request.topic
-        if not self.can_execute(identity):
```
lnielsen, line 41: Comment: Since the service already calls ``can_execute()`` I thought we wouldn't need it here.

invenio_requests/customizations/base/actions.py
```diff
@@ -29,35 +29,16 @@ def can_execute(self, identity, data=None):
         :param identity: The identity of the executor.
         :return: True if the action can be executed, False otherwise.
         """
-        return True
+        return self.status_from is None or \
```
lnielsen, line 32: Like this the action classes become a lot smaller

invenio_requests/customizations/base/schema.py
```diff
@@ -31,14 +31,14 @@ class RequestSchema(BaseRecordSchema):
     """Schema for requests."""
 
     number = fields.String(dump_only=True)
-    request_type = fields.String(dump_only=True)
-    title = SanitizedUnicode(required=True)
+    request_type = fields.String()
+    title = SanitizedUnicode(default='')
```
lnielsen, line 35: This we likely need to discuss. In RDM-Records I need a more loose schema to have a simpler REST API

invenio_requests/customizations/base/actions.py
```diff
@@ -29,35 +29,16 @@ def can_execute(self, identity, data=None):
...
+                self.request.status in self.status_from
 
-    def execute(self, identity, data=None):
+    def execute(self, identity, data=None, uow=None):
```
lnielsen, line 35: I'm thinking we should consider not passing in ``data`` (which right now is only used for the comment), so that we enforce a pattern where all data needed has to be contained in the request itself.

lnielsen:
For reference here's the RDM-Records stuff: https://github.com/inveniosoftware/invenio-rdm-records/pull/833/files

invenio_requests/customizations/base/actions.py
max-moser: true, as long as we control all the places where this method would be called (which is only in the service), we can make sure that `can_execute()` is always called before `execute()` (i.e. the implicit preconditions are met)

invenio_requests/customizations/base/actions.py
max-moser: then the statuses and actions really are like the states and transitions in automatons :nerd_face:

invenio_requests/customizations/default/actions.py
```diff
@@ -11,151 +11,44 @@
...
-            },
-        )
-        events_service.create(
-            identity,
-            request_id,
-            {
-                **data,
-                "type": RequestEventType.COMMENT.value,
-            },
-        )
```
max-moser, line 80: so the creation of comments and events would be moved outside of the `RequestAction` into the `service.execute_action()` method?

lnielsen: I would move the comment creation into the ``service.execution_action()`` and I would add the accepted/declined/... event by registering it on the Unit of Work:

```python
uow.register(EventOp(ActionEvent(self.status)))
```

Until that works, probably something like this in the ``RequestAction.execute()`` will do:
```
event_service.create(identity, request_id, ActionEvent(self.status))
```

invenio_requests/services/requests/service.py
```diff
@@ -214,7 +213,20 @@ def execute_action(self, identity, id_, action, data):
...
+
+        # current_events_service.create(
+        #     identity,
+        #     request.id,
+        #     {
+        #         **data,
+        #         "type": RequestEventType.COMMENT.value,
+        #     },
+        #     uow=uow,
+        # )
```
max-moser, line 229: wouldn't both things depend on the action being executed?:
* the type of event being created (same or analogous to the action type: "accept" -> `RequestEventType.ACCEPT`, ...)
* whether or not the event allows comments (yes for most, no for maybe submit and expire?)

so we could go with something like:
```python
# create event of associated type
action_obj = RequestActions.get_action(request, action)
event_type = action_obj.event_type  # the action knows what type of event it produces
current_events_service.create(identity, request.id, {**data, "type": event_type}, uow=uow)

# create comment if possible
if RequestActions.can_comment(identity, request, action):
    comment_type = RequestEventType.COMMENT.value
    current_events_service.create(identity, request.id, {**data, "type": comment_type}, uow=uow)
```

lnielsen: Yes, with one slight modification:

```python
# create event of associated type
action_obj = RequestActions.get_action(request, action)
event_type = action_obj.event_type  # the action knows what type of event it produces
if event_type is not None: #ADDED
    current_events_service.create(identity, request.id, {**data, "type": event_type}, uow=uow)

# create comment if possible
if RequestActions.can_comment(identity, request, action):
    comment_type = RequestEventType.COMMENT.value
    current_events_service.create(identity, request.id, {**data, "type": comment_type}, uow=uow)
```

max-moser (approved):
:+1:

invenio_requests/customizations/base/actions.py
```diff
@@ -8,56 +8,43 @@
...
 class RequestAction:
     """Base class for actions on requests."""
 
+    status_from = None
+    """Required status of a request to run this action."""
+
+    status_to = 'draft'
+    """Status after execution of the action."""
+
+    event_type = None
```
lnielsen, line 23: As discussed, now an action can define which event type it wants to log. By default nothing.

invenio_requests/customizations/base/actions.py
```diff
@@ -8,56 +8,43 @@
...
         This is mostly intended to be a hook for checking prerequisites
         (think, for instance, CI/CD pipeline runs).
         :param identity: The identity of the executor.
         :return: True if the action can be executed, False otherwise.
         """
-        return True
+        return self.status_from is None or \
+            self.request.status in self.status_from
 
-    def execute(self, identity, data=None):
```
lnielsen, line 34: ``data`` removed to enforce that actions are self-contained. In the service the data is only passed to the event service to create the comment.

invenio_requests/ext.py
```diff
@@ -9,9 +9,11 @@
 
 """Invenio module for generic and customizable requests."""
 
+import pkg_resources
 from invenio_base.utils import load_or_import_from_config
 
 from . import config
+from .registry import TypeRegistry
```
lnielsen, line 16: I created a general type registry which can be used for request types, entity resolvers and also event types (not done yet)

invenio_requests/ext.py
```diff
@@ -94,6 +95,23 @@ def init_resources(self):
...
             app.config["REQUESTS_REGISTERED_TYPES"]
         )
+        self.entity_resolvers_registry = TypeRegistry(
+            app.config["REQUESTS_ENTITY_RESOLVERS"]
+        )
+        # Load from entry points
+        register_entry_point(
+            self.request_type_registry,
+            'invenio_requests.types')
+        register_entry_point(
```
lnielsen, line 108: I couldn't find any nice easy way for RDM-Records/Communities to provide entity resolvers, so I went with entry point for now to move on.

invenio_requests/registry.py
```diff
@@ -0,0 +1,92 @@
...
+.. code-block:: python
+
+    registry.lookup("invenio-request.my-id")
+    registry.lookup_by_name("my_name")
+    for t in registry:
+        # do something
+"""
+
+
+class TypeRegistry:
```
lnielsen, line 31: Almost identical to the existing request type registry. Two additions:

- Each type can specify an alias/name and we can also look up by that name.
- You can iterate over the registered types (needed by the entity resolvers)

invenio_requests/resolvers/default/records.py
```diff
@@ -35,14 +35,15 @@ def get_need(self):
 class RecordResolver(EntityResolver):
     """Resolver for records."""
 
-    def __init__(self, record_cls, type_key="record"):
+    def __init__(self, record_cls, type_key="record", proxy_cls=RecordProxy):
```
lnielsen, line 38: In RDM-Records I needed to customize the proxy. The draft may or may not exists, so the resolver first looks for a record, then for a record.

invenio_requests/resolvers/default/requests.py
```diff
@@ -28,6 +28,8 @@ def get_need(self):
 class RequestResolver(EntityResolver):
     """Resolver for requests."""
 
+    type_id = 'request'
```
lnielsen, line 31: Needed for using the registry.

invenio_requests/services/events/service.py
```diff
@@ -40,20 +46,18 @@ def create(self, identity, request_id, data):
         record = self.record_cls.create(
             {},
             request=request.model,
-            request_id=request_id,
+            request_id=str(request_id),
```
lnielsen, line 49: I have to look further into this. The record is supposed to automatically dump this to a string, but if I don't do it here, it tries to stick a UUID object into the JSONSchema.

invenio_requests/services/events/service.py
```diff
@@ -172,7 +171,7 @@ def search(
             **kwargs,
         )
         if request_id:
-            search = search.filter("term", request_id=request_id)
+            search = search.filter("term", request_id=str(request.id))
```
lnielsen, line 174: Another thing we need to review is the user of the internal id vs the number and what is used for linking.

invenio_requests/services/requests/service.py
```diff
@@ -199,22 +199,40 @@ def rebuild_index(self, identity):
...
+        # Check if the action *can* be executed (i.e. correct status etc.)
+        if not action_obj.can_execute(identity):
             raise CannotExecuteActionError(action)
 
-        RequestActions.execute(identity, request, action, data)
+        # Execute action and register request for persistence.
+        action_obj.execute(identity, uow)
+        uow.register(RecordCommitOp(request, indexer=self.indexer))
+
+        # Create action event if defined
```
lnielsen, line 223: As discussed, now an action can decide if it wants to add log an event type.

invenio_requests/services/requests/service.py
```diff
@@ -199,22 +199,40 @@ def rebuild_index(self, identity):
...
+        event_type = action_obj.event_type
+        if event_type is not None:
+            current_events_service.create(
+                identity, request.id, {"type": event_type}, uow=uow
+            )
+
+        # Create comment if comment was provided
+        if data:
```
lnielsen, line 231: If data is provided, it's assumed you want to add a comment which is consistent with we've discussed in the REST API flow - i.e. all actions can have a comment added when being executed.

tests/services/events/test_request_events_service.py
```diff
@@ -31,7 +31,7 @@ def test_schemas(app, events_service_data, example_request):
 def test_simple_flow(app, identity_simple, events_service_data, example_request):
     """Interact with comment events."""
     events_service = current_requests.request_events_service
-    request_id = example_request.number
+    request_id = example_request.id
```
lnielsen, line 34: Related to the issue of internal vs external id (number)

invenio_requests/customizations/default/actions.py
lnielsen: Fixed in the updated PR.

invenio_requests/services/requests/service.py
lnielsen: Added in the updated PR

invenio_requests/ext.py
lnielsen: Here's an example from communities: https://github.com/inveniosoftware/invenio-communities/pull/364

invenio_requests/registry.py
```diff
@@ -0,0 +1,92 @@
...
+        for t in types:
+            self.register_type(t)
+
+    def register_type(self, type_, force=False):
+        """Register the specified request_type.
+
+        :param request_type: The request type to register.
+        :param as_type: Optional, the name under which to register the type.
+                        If not specified, the request_type name from the
+                        request_class will be taken.
```
max-moser, line 47: these params don't exist anymore

invenio_requests/registry.py
```diff
@@ -0,0 +1,92 @@
...
+            self._registered_names[type_name] = type_id
+        else:
+            self._registered_types.setdefault(type_id, type_)
+            self._registered_names.setdefault(type_name, type_id)
+
+    def lookup(self, type_id, quiet=False, default=None):
+        """Look up a registered type by its id.
+
+        :param type_id: The type name to look up.
+        :param quiet: TODO
```
max-moser, line 65: TODO
maybe let's omit the params and just add a longer textual description, as i've done in some other places?

max-moser (approved):
LGTM :+1:

PR #61, opened 2021-11-24
services: dynamic payload schemas

lnielsen:
With this change, a request type can only change the payload schema, and not the entire schema.


max-moser (approved):
LGTM
though maybe we should also adjust the mappings to not index the payload fields (and potentially crash and burn)?
https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/mappings/v7/requests/request-v1.0.0.json#L82-L85

PR #76, opened 2021-11-25
schemas: construct entity reference schema dynamically

### LINKED ISSUE #39: Marshmallow: Update marshmallow schema for EntityReferenceSchema
The `EntityReferenceSchema` should be updated to allow more than just `user` per default (but still only a single property should be accepted): https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/schema.py#L27

We need to investigate how to do that.
Maybe a `OneOfSchema` could help?
https://github.com/marshmallow-code/marshmallow-oneofschema

max-moser:
* only allow a single key to be present
* instead of allowing a fixed set of keys, create the set of accepted keys dynamically based on the request type, similar to the payload

closes #39

### DIFF
--- a/invenio_requests/customizations/default/request_types.py
+++ b/invenio_requests/customizations/default/request_types.py
@@ -77,13 +77,6 @@ class DefaultRequestType(RequestType):
     argument.
     """
 
-    marshmallow_schema = RequestSchema
-    """Schema used for de/serialization of requests of this type.
-
-    To be overridden in subclasses, if the custom request type follows a
-    different or more specific schema.
-    """
-
     def generate_request_number(self, request, **kwargs):
         """Generate a new request number.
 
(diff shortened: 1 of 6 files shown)


invenio_requests/customizations/default/request_types.py
```diff
@@ -77,13 +77,6 @@ class DefaultRequestType(RequestType):
     argument.
     """
 
-    marshmallow_schema = RequestSchema
-    """Schema used for de/serialization of requests of this type.
-
-    To be overridden in subclasses, if the custom request type follows a
-    different or more specific schema.
-    """
-
```
max-moser, line 86: this shadowed the `marshmallow_schema` property of the base class – otherwise, the tests wouldn't have passed because we often pass a `RequestType` class (rather than an instance) into the `service.create(…)`, which would fail with a `TypeError: 'property' object is not callable` because properties can't be used from classes

lnielsen (approved):

PR #78, opened 2021-11-26
Fixes for Invenio-RDM-Records

lnielsen:


invenio_requests/customizations/base/request_types.py
```diff
@@ -140,18 +139,20 @@ def marshmallow_schema(cls):
...
         """
-        return str(uuid4())
+        from invenio_requests.records.models import RequestNumber
+        return base32.encode(RequestNumber.next())
 
     def __str__(self):
         """Return str(self)."""
-        return self.name
+        # Value used by marshmallow schemas to represent the type.
+        return self.type_id
```
lnielsen, line 154: Changed this so that Marshamllow serialises the ``type_id`` instead of the ``name``.

invenio_requests/customizations/default/request_types.py
```diff
@@ -76,12 +76,3 @@ class DefaultRequestType(RequestType):
     RequestAction class is instantiated with the current Request object as
     argument.
     """
-
```
lnielsen, line 79: Defined in the BaseRequestType

invenio_requests/records/api.py
```diff
@@ -59,7 +58,7 @@ class Request(Record):
     schema = ConstantField("$schema", "local://requests/request-v1.0.0.json")
     """The JSON Schema to use for validation."""
 
-    type = RequestTypeField("request_type_id")
+    type = RequestTypeField("type")
```
lnielsen, line 61: I've aligned it so that the JSON stored, index and marshmallow dumped are all using the same key (``type``).

invenio_requests/records/api.py
```diff
@@ -42,7 +42,6 @@ class Request(Record):
     dumper = ElasticsearchDumper(
         extensions=[
             CalculatedFieldDumperExt("is_open"),
-            RequestTypeDumperExt("type"),
```
lnielsen, line 45: This isn't needed anymore when we're just dumping the ``type_id`` as it's the same as in the DB.

invenio_requests/services/requests/results.py
```diff
@@ -37,7 +37,7 @@ def __init__(
     @property
     def id(self):
         """Identity of the request."""
-        return self._request.number
+        return str(self._request.id)
```
lnielsen, line 40: I've' not changed the ``Request.get_record()`` but potentially, if we use the uuid as id, and ``number`` is more like metadata, then probably we won't need a custom ``get_record()``

invenio_requests/customizations/base/request_types.py
```diff
@@ -140,18 +139,20 @@ def marshmallow_schema(cls):
...
-        """Generate a new external identifier.
+        """Generate a new request number identifier.
 
         This method can be overridden in subclasses to create external identifiers
         according to a custom schema, using the information associated with the request
         (e.g. topic, receiver, creator).
         """
-        return str(uuid4())
+        from invenio_requests.records.models import RequestNumber
+        return base32.encode(RequestNumber.next())
```
lnielsen, line 149: Basic start for now.

invenio_requests/services/requests/results.py
max-moser: i.e. we wouldn't use it to look requests up, but only use it for display purposes?
in this case, we wouldn't need any uniqueness/identifying properties(*) on it, and could just generate a sequential number on per-receiver basis?
and we could also drop the database column, and put it into the request's json?

(*) not entirely true, we would still need to ensure that the number is unique on a per-receiver basis, to avoid confusion (e.g. having two requests number 1 for the same community)

invenio_requests/records/api.py
max-moser: so previously we had `request_type_id` (the ID of the type), and `type` (the name of the type) stored
and now we dropped storing the name (because it's synthesized by the `RequestType` classes), and only store the ID anymore (under the field `type`)?

max-moser (approved):

invenio_requests/records/api.py
lnielsen: Yes, because we can always (in code) get the name by knowing the ID (``request.type.name``).

PR #102, opened 2021-12-17
services: refactor configuration

fenekku:
A simple refactoring of the custom configuration pattern. See invenio-rdm-records for a more extensive one: <https://github.com/inveniosoftware/invenio-rdm-records/pull/865>

The goal is to place code that deals with configuration together and get rid of intermediate cruft.


max-moser (approved):
LGTM; WDYT @lnielsen?

fenekku:
Shifted to passing `app` as an argument to `Configurator.build` and storing it on the config class. The stored `app.config` is then ultimately used for configuration resolution.

max-moser:
so now the `Configurator.build()` also has a definite purpose!

PR #104, opened 2022-01-20
api: only have get_record resolve requests by id

### LINKED ISSUE #74: Don't have Request get_record interpret its id in 2 ways
`get_record` takes an `id` and checks if there is a row with matching `external_id` and if not, if there is a row with matching id: https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/records/api.py#L107 . But since `external_id` and `id` are populated with UUIDs right now we may end up with a erroneous get e.g. record A has as external_id the same id value as record B. 

I think this was a "it-would-be-nice-if" but it's not nice :) . Let's make different gets for different ids. It's fine.

max-moser:
closes #74 

* actually, it is always the request ID that's used for external interfaces to interact with requests, and not the number
* thus, there was no real benefit in having Request.get_record resolve by either ID or number
* this only opened up the possibility of having one request's number shadow another request's ID and thus, potential issues

### DIFF
--- a/invenio_requests/records/api.py
+++ b/invenio_requests/records/api.py
@@ -95,38 +95,6 @@ class Request(Record):
     is_expired = ExpiredStateCalculatedField("expires_at")
     """Whether or not the request is already expired."""
 
-    @classmethod
-    def get_record(cls, id_, with_deleted=False):
-        """Retrieve the request by id.
-
-        :param id_: The record's number or internal ID.
-        :param with_deleted: If `True`, then it includes deleted requests.
-        :returns: The :class:`Request` instance.
-        """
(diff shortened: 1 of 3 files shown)


invenio_requests/records/api.py
```diff
@@ -95,38 +95,6 @@ class Request(Record):
...
-                query = query.filter(cls.model_cls.is_deleted != True)  # noqa
-
-            model = query.one()
-
-        if model is None:
-            # TODO maybe some kind of `NullRequest`?
-            return None
-
-        return cls(model.data, model=model)
-
```
fenekku, line 129: :firecracker:

fenekku (approved):

PR #113, opened 2022-01-26
Allow override of action permissions per request type

max-moser:
We may want to be able to customize the action permissions for each request type specifically, e.g. change the required permissions to perform a specific action on a particular request type, or just disable said action for the given request type on your own Invenio instance overall.

# Example
We want the default behavior for the **accept** action (available to the **receiver(s)** and the **system**).
However, we want to disable the action of the same name for requests of type FooBar (`request.type.type_id = "foo-bar"`).

This could be accomplished with the following permission policies:
```python
can_action_accept = [Receiver(check=is_open), SystemProcess()]
can_action_accept_foo_bar = [Disable()]
```

# Alternative Approach
One alternative approach would be to have optional sub-permission-policies per request type, defining their permission overrides, e.g.:
```python
class FooBarPermissionPolicy(RecordPermissionPolicy):
  can_action_accept = [Disable()]
  # ...

class RequestPermissionPolicy(RecordPermissionPolicy):
  can_action_accept = [Receiver(check=is_open), SystemProcess()]
  # ...
  override_foo_bar = FooBarPermissionPolicy()
```

### DIFF
--- a/invenio_requests/services/events/service.py
+++ b/invenio_requests/services/events/service.py
@@ -19,6 +19,7 @@ from invenio_records_resources.services.uow import (
 )
 
 from ...records.api import RequestEventType
+from ..permissions import get_most_specific_policy_name
 
 
 class RequestEventsService(RecordService):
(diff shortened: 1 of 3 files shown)


max-moser:
If it is decided to go with another approach, we should not forget to remove the currently existing stub for this approach: https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/services/requests/service.py#L34-L49

If this is accepted, I will add some test cases before merging it.

max-moser:
closed in favor of https://github.com/inveniosoftware/invenio-requests/pull/121 which simplifies the permission look-ups rather than expands them, as we decided that we will only add this feature if the need arises.

PR #116, opened 2022-01-28
ui: rearranged structure

### LINKED ISSUE #107: Request detail: general structure
upload_6a0577e676984aca53ceb2bd48f8dfef

RFC:
https://codimd.web.cern.ch/ICyKqzFLQgi3PXAnMcMxlA
Decisions/discussion points:
https://codimd.web.cern.ch/yxBgEdVITVqgGr44RnF3Hw

kprzerwa:
* added loader
* added ErrorBoundary
* closes #107


invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+const requestDetailsDiv = document.getElementById("request-detail");
+const overriddenCmps = {};
+
+ReactDOM.render(
+  <OverridableContext.Provider value={overriddenCmps}>
+    <RequestDetails request={JSON.parse(requestDetailsDiv.dataset.record)} />
+  </OverridableContext.Provider>,
+
+  requestDetailsDiv
```
zzacharo, line 14: EXtra space?

invenio_requests/assets/semantic-ui/js/invenio_requests/requestDetails.js
```diff
@@ -1,11 +1,9 @@
 import React, { Component } from "react";
```
zzacharo, line 1: We are not using camelcase names for the component files? E.g https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/RecordManagement.js

invenio_requests/assets/semantic-ui/js/invenio_requests/requestDetails.js
```diff
@@ -41,14 +38,7 @@ class RequestDetails extends Component {
 }
 
 RequestDetails.propTypes = {
-  request: PropTypes.object.isRequired,
+  request: PropTypes.object.isRequired
```
zzacharo, line 41: it would be nice when this is finalized to convert the prop type in a `PropTypes.shape({...})`

Pineirin (approved):
LGTM.
One question. Should we make every component of the application overridable or only a certain set of them?

invenio_requests/views/decorators.py
```diff
@@ -25,11 +26,31 @@ def pass_request(f):
...
+    def view(**kwargs):
+        """."""
+        request = kwargs.get("request")
+        if request is None:
+            # avoid running service twice with stacked decorators
+            pid_value = kwargs['pid_value']
+            request = service().read(
+                id_=pid_value, identity=g.identity
+            )
+        if request.has_permissions_to(["read"])["can_read"]:
```
zzacharo, line 51: is the `has_permissions_to()` not returning a true or false value?

invenio_requests/assets/semantic-ui/js/invenio_requests/requestDetails.js
kprzerwa: it is a camelCase name of the filename. You probably mean PascalCase.
personally I am not a big fan of PascalCase, but if it a standard or agreement in RDM, I will follow it. Is it something that was agreed on, a pattern ? should it be discussed?

invenio_requests/assets/semantic-ui/js/invenio_requests/requestDetails.js
kprzerwa: I think it highly depends on the final serializers, which the other team is working on, so I think it is too soon to do that at this stage, I can add it to my TODOs list

invenio_requests/views/decorators.py
kprzerwa: actually the service is checking already for permissions, so there is no reason for this decorator to exist, I will remove it.
to answer: yes, it returns True/False, but I am afraid I am missing the context of your question. Could you elaborate ?

invenio_requests/assets/semantic-ui/js/invenio_requests/api/config.js
```diff
@@ -0,0 +1,40 @@
+// This file is part of Invenio-requests
+// Copyright (C) 2022 CERN.
+//
+// Invenio-requests is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import axios from "axios";
```
ntarocco, line 7: Good! We should discuss moving this to a more "base/core" module.
If we also add UI cmps in such module, I would propose to evaluate the integration of [Storybook](https://storybook.js.org/docs/react/get-started/introduction)

invenio_requests/assets/semantic-ui/js/invenio_requests/components/Loader.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from 'react';
+import Overridable from 'react-overridable';
+import { Loader as UILoader } from 'semantic-ui-react';
+import PropTypes from 'prop-types';
+
+class Loader extends Component {
+  render() {
+    const { isLoading, children } = this.props;
+    return (
+      <Overridable id="Loader.layout" {...this.props}>
```
ntarocco, line 10: Isn'it this `id` too generic? Shouldn't we namespace it by "module", e.g. `RequestDetails.Loader.layour`?
I am wondering what happens if I have the same cmp in another details page.
Same for the other `ids`.

invenio_requests/views/decorators.py
```diff
@@ -25,11 +26,31 @@ def pass_request(f):
...
         return f(**kwargs)
 
     return view
+
+
+def require_request_permissions(f):
+    """Decorate a view to require request read accessing the view."""
+    @wraps(f)
+    def view(**kwargs):
+        """."""
```
ntarocco: Doc string to be added here.

invenio_requests/views/decorators.py
```diff
@@ -25,11 +26,31 @@ def pass_request(f):
...
+        request = kwargs.get("request")
+        if request is None:
+            # avoid running service twice with stacked decorators
+            pid_value = kwargs['pid_value']
```
ntarocco: move this to a common private func to avoid code duplication?

invenio_requests/assets/semantic-ui/js/invenio_requests/requestDetails.js
zzacharo: @kprzerwa yes sorry PascalCase. We have named everything following that convention and personally my rule is that:
1. If a js file is defining a component and exporting it, then I am using PascalCase,
2. camelCase in all other cases

I think at the moment there is a big number of js files following the above convention, including https://github.com/inveniosoftware/react-invenio-deposit/tree/master/src/lib, so it would be impossible to change it now!

invenio_requests/assets/semantic-ui/js/invenio_requests/requestDetails.js
zzacharo: Yes sorry,  I definitely meant for later down the road! Noting it down is fine be me!

invenio_requests/views/decorators.py
zzacharo: I found really weird given the name i.e `request.has_permissions_to(["read"])` to check `["can_read"]` afterwards. But better if you remove it :) Otherwise, I might had hide the implementation detail behind another function e.g `can_user_view_request`.

invenio_requests/webpack.py
```diff
@@ -44,9 +44,8 @@
                 'qs': '^6.9.1',
                 'react': '^16.12.0',
                 'react-dom': '^16.11.0',
-                'redux': '^4.0.5',
-                'redux-thunk': '^2.3.0',
                 'yup': '^0.27.0',
+                'react-overridable': '^0.0.3',
```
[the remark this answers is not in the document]
zzacharo: Well, that might be a bigger task as we need to check react-invenio-deposit and the compatibility with the latest semantic-ui-react. Definitely needs to be done but we might want to make an issue for it at the moment. WDYT?

kprzerwa:
> LGTM. One question. Should we make every component of the application overridable or only a certain set of them?

@Pineirin I think it depends on the use case, where we want to allow for overridability as developers. In this case I am still expanding the set of the overridables and in the next PRs you will see more of them

invenio_requests/assets/semantic-ui/js/invenio_requests/RequestDetails.js
```diff
@@ -1,15 +1,12 @@
...
-import Timeline from "./timeline";
-
-const requestDetailsDiv = document.getElementById("request-detail");
+import Timeline from "./Timeline";
 
 class RequestDetails extends Component {
   get menuPanes() {
-    const { request } = this.props;
-    console.log(request);
+    const { request, api } = this.props;
```
ntarocco, line 9: Missing from `propTypes`.

invenio_requests/assets/semantic-ui/js/invenio_requests/Timeline.js
```diff
@@ -0,0 +1,62 @@
...
+
+  fetchData = async () => {
+    const { api } = this.props;
+    try {
+      this.cancellableFetchData = withCancel(api.getTimeline());
+      const response = await this.cancellableFetchData.promise;
+      this.setState({ timelineData: response.data, loading: false });
+    } catch (error) {
+      if (error !== "UNMOUNTED") {
+        console.error(error);
```
ntarocco, line 32: what if we move this to the API layer to avoid to copy/paste it in each try/catch?

invenio_requests/assets/semantic-ui/js/invenio_requests/Timeline.js
```diff
@@ -0,0 +1,62 @@
...
+  render() {
+    const { timelineData, loading } = this.state;
+    return (
+      <Overridable id="InvenioRequests.Timeline.layout" {...this.props}>
+        <Container>
+          <Loader isLoading={loading}>
+            <Segment>
+              <Feed>
+                {timelineData.hits?.hits.map((comment) => (
+                  <TimelineEvent event={comment} key={comment.id} />
```
ntarocco, line 47: Probably, in the future, this should become a selector of cmp depending on the type? It could be a Comment, or a Message, or anything else, and specialized cmp should be used instead.

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
```diff
@@ -0,0 +1,24 @@
...
+import RequestDetails from './RequestDetails';
+
+const requestDetailsDiv = document.getElementById("request-detail");
+const overriddenCmps = {};
+const resource = JSON.parse(requestDetailsDiv.dataset.record);
+const api = new InvenioRequestsAPI(new RequestLinkExtractor(resource.links));
+
+ReactDOM.render(
+  <OverridableContext.Provider value={overriddenCmps}>
+    <RequestDetails request={resource} api={api}/>
```
ntarocco, line 16: Why passing `api` here? I think it is because you are using the constructor to pass `resource.links` obj.
In ES6, you cannot recreate a new obj when you import, but with my proposed solution you could create an `init` method that will set it (instead of the constructor).
In this way, you don't have to pass it around: every time you import it, you should have the same instance.

invenio_requests/assets/semantic-ui/js/invenio_requests/TimelineEvent.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from "react";
+import { Feed, Image } from "semantic-ui-react";
+import PropTypes from "prop-types";
+import Overridable from "react-overridable";
+
+const TimelineEvent = ({ event }) => {
+  // TODO i18n
```
zzacharo, line 7: @kprzerwa maybe open a ticket withj i18n integation just not to forget and then you remove this TODO!

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
```diff
@@ -0,0 +1,24 @@
...
+   </OverridableContext.Provider>,
+  requestDetailsDiv
+);
+
+export { default as RequestDetails } from "./RequestDetails";
+export { default as Timeline } from "./Timeline";
+export { default as TimelineEvent } from "./TimelineEvent";
+export { InvenioRequestsAPI, RequestLinkExtractor } from "./api";
```
ntarocco, line 24: Why exporting these? They should not be reused?

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
ntarocco, line 16: A little warning: I would suggest to stick to one naming only. Here we have `dataset.record`, which becomes `resource`, which becomes `request`.
In ILS we had a similar issue: hard to say  if the prop we have passed, contains the `metadata` field, or it is already the metadata.
E.g. `record`, when passing the entire record, and `recordMetadata` when passing `metadata`. Or maybe `request` and `requestMetadata`.

invenio_requests/assets/semantic-ui/js/invenio_requests/api/api.js
```diff
@@ -0,0 +1,43 @@
...
+    if (!this.links.actions) {
+      throw TypeError("Actions link missing from resource.");
+    }
+    return this.links.actions;
+  };
+}
+
+export class InvenioRequestsAPI {
+  constructor(apiLinks) {
+    this.links = apiLinks;
```
zzacharo, line 35: What about adding renaming`apiLinks` to `requestLinkExtractor` ? In the beginning I had a comment about changing the parameter to use the request link extractor because I thought that the apiLinks is just an object with links...

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
zzacharo: I will add just that if the api is needed in not just the direct children of `RequestDetails` cmp, then we will need to pass it to the context/redux!

invenio_requests/assets/semantic-ui/js/invenio_requests/TimelineEvent.js
kprzerwa: there is already one

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
kprzerwa: I thought it would be useful for overriding, especially if you want to override the whole component. I will remove the link extractor class but I would leave the API. what do you think ?

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
kprzerwa: I wanted to make the component reusable in the sense that it is not dependable on the InvenioRequestAPI, redux store nor context. It was mainly meant for the Timeline

invenio_requests/assets/semantic-ui/js/invenio_requests/Timeline.js
kprzerwa: yes that is correct, I planned to work on this part later when I focus on the timeline, not general structure

invenio_requests/assets/semantic-ui/js/invenio_requests/Timeline.js
kprzerwa: correct me if I am wrong but if I don't check if the error was `UNMOUNTED` I will still have a memory leak on setting state, no?

invenio_requests/assets/semantic-ui/js/invenio_requests/index.js
ntarocco: Having a reusable UI cmp getting as a `prop` the backend request adapter is, IMO, "dangerous" and not flexible: the UI cmp will have to know exactly what method to call in the adapter and what is the response format.
The reusable cmp should, instead, simply access the already fetched `events` in a specific format.
An upper non-reusable cmp should:
1. fetch the data via the specific backend adapter
2. inject the data to the reusable cmp responsible for displaying it.

This allows you to decouple the 2. Having the API injected, it is actually coupling it. Happy to discuss this in IRL.

PR #127, opened 2022-02-04
Service: add facets

jrcastro2:
* resource: add is_open param

### DIFF
--- a/invenio_requests/services/requests/config.py
+++ b/invenio_requests/services/requests/config.py
@@ -13,6 +13,8 @@
 from invenio_records_resources.services import RecordServiceConfig, SearchOptions
 from invenio_records_resources.services.records.links import pagination_links
 
+from invenio_requests.services.requests import facets
+
 from ...customizations.base import RequestActions
 from ...records.api import Request
 from ..configurator import ConfiguratorMixin, FromConfig
@@ -24,7 +26,7 @@ from .components import (
     RequestNumberComponent,
 )
 from .links import RequestLink
-from .params import ReferenceFilterParam
+from .params import IsOpenParam, ReferenceFilterParam
 from .results import RequestItem, RequestList
 
 
@@ -42,7 +44,13 @@ class RequestSearchOptions(SearchOptions):
         ReferenceFilterParam.factory(param="created_by", field="created_by"),
         ReferenceFilterParam.factory(param="receiver", field="receiver"),
         ReferenceFilterParam.factory(param="topic", field="topic"),
+        IsOpenParam.factory("is_open"),
     ]
+
+    facets = {
+        'type': facets.type,
+        'status': facets.status,
+    }
 
 
 class RequestsServiceConfig(RecordServiceConfig, ConfiguratorMixin):
--- a/invenio_requests/services/requests/facets.py
+++ b/invenio_requests/services/requests/facets.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Facet definitions."""
+
+from flask_babelex import gettext as _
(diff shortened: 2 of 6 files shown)


invenio_requests/services/requests/config.py
```diff
@@ -24,10 +24,9 @@
 )
 from .customization import RequestsConfigMixin
 from .links import RequestLink
-from .params import ReferenceFilterParam
+from .params import ReferenceFilterParam, IsOpenParam
 from .results import RequestItem, RequestList
-
-
+from invenio_rdm_records.services import facets
```
zzacharo, line 29: I am checking invenio-rdm-records but I don't see the `type` and `status` facet defined there....I might be blind though 😅  Anyway, isn't the type facet related to the request type rather than the record type? I.e shouldn't we define the facet here?

jrcastro2: I forgot to create the PR for invenio-rdm-records with the facets so you cannot finde them (sorry). I thought that we would define all the facets for all the records in rdm-records, but it makes sense to have the facets defined here in invenio-requests. Thanks!

jrcastro2: I added the facets here and removed them from invenio-rdm-records but by doing so I changed it to merge 2 dicts when creating the searchApp (check [here](https://github.com/inveniosoftware/invenio-app-rdm/pull/1212/files#diff-3ba66c56d880d75a9fd4cef92b5f5bc07ebe6dcbd28461afd900246f88ee2da7R207)), this can cause issues if we have 2 facets in different records but with the same name, I think it should be ok as we control it. Alternatively we can pass the facets config on creating the searchApp and just get the facets for each record instead of merging all of them. Let me know if I should change it to avoid the merging.

zzacharo (approved):

invenio_requests/services/requests/params.py
```diff
@@ -53,3 +55,25 @@ def apply(self, identity, search, params):
...
+        """Create a new filter parameter."""
+        return partial(cls, field)
+
+    def apply(self, identity, search, params):
+        """Evaluate the is_open parameter on the search."""
+        if params.get("is_open") is True:
+            search = search.filter('term', **{self.field_name: True})
+        elif params.get("is_open") is False:
+            search = search.filter('term', **{self.field_name: False})
+        return search
```
ntarocco, line 79: Should we raise a `400` if not `[True|False]`?

zzacharo: Do we want to fail the whole search if a param is invalid? At the moment, it just not taken into account, which is possibly bad if there is a mistake but at least it doesn't fail the whole query...We did something similar in [allversions](https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/search_params.py#L33) search parameter. In that case we are just checking for the existence or not...maybe we could do that all the requests are either closed or open!

invenio_requests/services/requests/facets.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020-2021 CERN.
```
zzacharo, line 3: 2022

invenio_requests/services/requests/params.py
jrcastro2: From the UI we can only select `Open` or `Closed` and we could just allow one or another but if we do this we do not allow to retrieve all the requests (closed and open) this would also solve this issue; https://github.com/inveniosoftware/invenio-app-rdm/issues/1214. That said I would probably avoid it unless we know that we don't want to retrieve all the requests in the future.

About the 400, I don't have a strong opinion but I understand that we should align with other parameters and as Zach mentioned at the moment is not failing, just ignoring it.

invenio_requests/services/requests/config.py
jrcastro2: After discussion we decided to pass only the available facets to each SearchApp instead of mergind all dicts, so no issue should arise

kpsherva (commented):
please change the destination branch

PR #136, opened 2022-02-11
request types custom permission policy

fenekku:
- requests: customize permission policy per type
- release v0.2.5


fenekku:
Talked about it and addressed comments. Per discussion earlier in the week, I am merging these PRs and we can always go back if we have an epiphany.

PR #140, opened 2022-02-16
style: minor fixes

jrcastro2:


invenio_requests/assets/semantic-ui/less/invenio_requests/theme.less
```diff
@@ -14,3 +14,6 @@
   }
 }
 
+.request-detail-page {
```
kpsherva, line 17: why not to use the margins classes on the element?

kpsherva (approved):

PR #146, opened 2022-02-21
ui: add license information

kpsherva:


invenio_requests/assets/semantic-ui/js/invenio_requests/components/Error.js
```diff
@@ -1,3 +1,4 @@
+
```
jrcastro2, line 1: License missing here

ntarocco (approved):
Checked very quickly, go ahead and merge ;)

invenio_requests/assets/semantic-ui/js/invenio_requests/components/Error.js
kpsherva: I plan to move this component to a common module

PR #151, opened 2022-02-23
Disentangle states and permissions

### LINKED ISSUE #122: Refactor permissions, can_execute and link generation
Currently to generate links for actions, we call ``action.can_execute()``. For the community submission the ``can_execute()`` method on the accept actions validates the draft. Since this is called on every request on a search result, this could lead to a big performance penalty.

Overall, this issue has highlighted some issues wrt responsibilities. Following came up under discussion with @max-moser.

### ```Action.can_execute()`` should only deal with state transitions.

``Action.can_execute()`` should only check if the request is in a state that's allowed.

### Permissions should only consider the creator and receiver.

Permission to perform a given action should rely only on who is the creator and receiver of a request. Right now the request state is sometimes taken into account when a permission is checked. This makes it hard to reason about and understand the permissions.

### Request states

Given above two principles, we have conceptual model like below:

Screenshot 2022-02-03 at 16 00 06

Thus, we'll need to e.g. split the ``can_update`` action into two different actions (update draft, and update) because it's dealing with two different request states.

lnielsen:
Closes #122 

Depends on inveniosoftware/invenio-records-resources#303


invenio_requests/services/events/service.py
```diff
@@ -90,108 +74,76 @@ def read(self, identity, id_):
...
-        record = self._get_event(id_)
-        request = self._get_request(record.request.id)
-        # this service method doesn't allow type change
-        data["type"] = record.type
-
-        self.check_revision_id(record, revision_id)
+        """Update a comment (only comments can be updated)."""
+        event = self._get_event(id_)
+        request = self._get_request(event.request.id)
+        self.require_permission(
```
ntarocco, line 80: Does it make sense to move this line, `require_permission`, after `if event.type != RequestEventType.COMMENT.value:`? Or maybe move the check that the event type is a comment as first thing?

After that `if`, we know that the user is actually changing a comment, and not an event.

invenio_requests/services/events/service.py
```diff
@@ -90,108 +74,76 @@ def read(self, identity, id_):
...
+            identity, "delete_comment", request=request, event=event)
+        self.check_revision_id(event, revision_id)
 
-        We may want to add a parameter to assert a kind of event is deleted
-        to prevent the weird semantic of using the comments REST API to
-        delete an event (which is only possible for an admin anyway).
-        """
-        record = self._get_event(id_)
-        request = self._get_request(record.request_id)
+        if event.type != RequestEventType.COMMENT.value:
```
ntarocco, line 117: same here

invenio_requests/services/requests/config.py
```diff
@@ -34,7 +33,10 @@ def _is_action_available(request, context):
     """Check if the given action is available on the request."""
     action = context.get("action")
     identity = context.get("identity")
-    return RequestActions.can_execute(identity, request, action)
+    permission_policy_cls = context.get('permission_policy_cls')
```
ntarocco, line 36: `    permission_policy_cls = context['permission_policy_cls']` and maybe the same for `identity` and `action`?

invenio_requests/services/generators.py
```diff
@@ -0,0 +1,98 @@
...
+    """Generator to validate needs only for a given request status."""
+
+    def __init__(self, statuses, generators):
+        """Initialize need entity generator."""
+        self._statuses = statuses
+        self._generators = generators or []
+
+    def needs(self, request=None, **kwargs):
+        """Needs if status is in one of the provided ones."""
+        if request.status in self._statuses:
```
ntarocco, line 32: can `request` param be `None`?

invenio_requests/services/events/service.py
lnielsen: Yes 👍

invenio_requests/services/generators.py
lnielsen: I still have to test, but in principle it should not be possible. It's was a ``if record is None: return`` that caused the recent vulnerability, so was trying to avoid adding it here, but perhaps I should simply add a ``assert request is not None`` to ensure it's always set.

invenio_requests/services/requests/components.py
```diff
@@ -38,26 +38,11 @@ class RequestDataComponent(DataComponent):
...
-        record.clear_none()
-
-
-class DefaultStatusComponent(ServiceComponent):
-    """Component for initializing the default status of the request."""
-
-    def create(self, identity, data=None, record=None, **kwargs):
-        """Initialize the default status of the request."""
-        record.status = record.type.default_status
+        if record.status == 'created':
```
lnielsen, line 41: This part still needs to be checked in detail which properties we should be able to change. Previously the code had a bug here that could allow it to e.g. unset the status.

invenio_requests/services/requests/service.py
```diff
@@ -104,6 +89,9 @@ def create(
             uow=uow,
         )
 
+        # Get and run the request type's create action.
+        self._execute(identity, request, request_type.create_action, uow)
```
lnielsen, line 93: This allow a request to execute a create action (i.e. this action could setup buckets and similar for uploading files).

invenio_requests/services/requests/service.py
```diff
@@ -180,17 +164,16 @@ def delete(self, identity, id_, uow=None):
...
-
-        # TODO:
-        # prevent deletion if in open state?
+        self.require_permission(identity, f"delete", request=request)
 
         # run components
         self.run_components("delete", identity, record=request, uow=uow)
 
+        # Get and run the request type's create action.
+        self._execute(identity, request, request.type.delete_action, uow)
```
lnielsen, line 173: Similarly here, we allow a delete action which can clean up from what was created.

invenio_requests/customizations/actions.py
```diff
@@ -27,16 +27,18 @@ def __init__(self, request):
...
-        (think, for instance, CI/CD pipeline runs).
-        :param identity: The identity of the executor.
+        This checks if a given state transition for a request is allowed or
+        not.
+
         :return: True if the action can be executed, False otherwise.
         """
-        return self.status_from is None or \
-            self.request.status in self.status_from
+        if self.status_from is None:
```
lnielsen, line 38: A request is by default created with status ``None``. The create action is responsible for transitioning into the next state.

invenio_requests/customizations/__init__.py
```diff
@@ -1,20 +1,38 @@
...
+    CancelAction,
+    CreateAction,
+    DeclineAction,
+    DeleteAction,
+    ExpireAction,
+    RequestAction,
+    RequestActions,
+    SubmitAction,
+)
+from .request_types import RequestType
```
lnielsen, line 22: I have simplified the code to a single request type instead of ``BaseRequestType`` and ``DefaultRequestType``. This was possible by removing the possibility for an action to defined an event type that should be logged. This code should instead be instead moved up to the service (or into the specific request type)

invenio_requests/customizations/request_types.py
```diff
@@ -43,21 +51,44 @@ class RequestType:
     name = "Base Request"
     """The human-readable name for this type of requests."""
 
-    available_statuses = {}
-    """Available statuses for the Request.
+    available_statuses = {
+        "created": RequestState.UNDEFINED,
```
lnielsen, line 55: Renamed from ``NEITHER`` to ``UNDEFINED``.

invenio_requests/customizations/request_types.py
```diff
@@ -98,7 +129,20 @@ class RequestType:
         }
     """
 
-    permission_policy_cls = BaseRequestPermissionPolicy
+    needs_context = None
```
lnielsen, line 132: The key change which allow a request type to customise the needs generated for a creator or receiver. Basically the ``get_needs()`` of an entity reference can be passed in a context object, that allows the ``get_needs()`` method to behave differently depending not he context (e.g. generate different community roles when the entity is a community).

invenio_requests/records/systemfields/status.py
```diff
@@ -28,9 +28,4 @@ def __get__(self, record, owner=None):
             # access by class
             return self
 
-        status = self.get_dictkey(record)
-        if status is None:
```
lnielsen, line 32: Allow a request status to be none, and make sure it's the actions responsibility to set the new status (i.e. state transition)

invenio_requests/services/events/config.py
```diff
@@ -48,10 +48,6 @@ class RequestEventsServiceConfig(RecordServiceConfig, ConfiguratorMixin):
     )
     schema = RequestEventSchema
     record_cls = RequestEvent
-    components = [
-        DataComponent,
```
lnielsen, line 52: Moved it into the service instead to simplify, and we don't have a need to customize here.

invenio_requests/services/events/service.py
```diff
@@ -46,45 +30,31 @@ def create(self, identity, request_id, data, uow=None):
         :param dict data: Input data according to the data schema.
         """
         request = self._get_request(request_id)
-        permission = self._get_permission("create", data["type"])
-        self.require_permission(
-            identity, permission, request_type=request.type, request=request
-        )
+        # If you can read the request you can create events for the request.
+        self.require_permission(identity, "read", request=request)
```
lnielsen, line 34: Simplified permissions to reply on the request's ``read`` method (if you can read an event, you can comment on it)

invenio_requests/services/events/service.py
```diff
@@ -93,10 +63,7 @@ def read(self, identity, id_):
         record = self._get_event(id_)
         request = self._get_request(record.request_id)
 
-        # Same "read_event" permission for all types of events
-        self.require_permission(
-            identity, "read_event", request_type=request.type, request=request
-        )
+        self.require_permission(identity, "read", request=request)
```
lnielsen, line 66: Using ``read`` because it needs to take request status into account.

invenio_requests/services/events/service.py
```diff
@@ -107,115 +74,76 @@ def read(self, identity, id_):
...
+        """Update a comment (only comments can be updated)."""
+        event = self._get_event(id_)
+        request = self._get_request(event.request.id)
         self.require_permission(
-            identity, permission, request_type=request.type, event=record
-        )
+            identity, "update_comment", request=request, event=event)
+        self.check_revision_id(event, revision_id)
 
+        if event.type != RequestEventType.COMMENT.value:
```
lnielsen, line 84: We should only allow updating comments. Other types of events we should not allow updating. If we need for some reason to update them, I think we should split into two methods.

invenio_requests/services/events/service.py
```diff
@@ -107,115 +74,76 @@ def read(self, identity, id_):
...
-
-        We may want to add a parameter to assert a kind of event is deleted
-        to prevent the weird semantic of using the comments REST API to
-        delete an event (which is only possible for an admin anyway).
-        """
-        record = self._get_event(id_)
-        request = self._get_request(record.request_id)
-
-        self.check_revision_id(record, revision_id)
+        """Delete a a comment (only comments can be deleted)."""
```
lnielsen, line 108: Only allow "deleting" comments.

invenio_requests/services/events/service.py
```diff
@@ -107,115 +74,76 @@ def read(self, identity, id_):
...
-               **kwargs):
-        """Search for events matching the querystring.
+        uow.register(
+            RecordCommitOp(event, indexer=self.indexer)
+        )
+        return True
 
-        Request_id is optional.
-        """
+    def search(self, identity, request_id, params=None, es_preference=None, **kwargs):
```
lnielsen, line 128: Simplified search to always require searching over a request. This simplifies application of correct permissions. If we need global event search, then I'd prefer a separate search method, that's probably for admins only.

invenio_requests/services/events/service.py
```diff
@@ -107,115 +74,76 @@ def read(self, identity, id_):
...
+        self.require_permission(identity, "read", request=request)
 
         # Prepare and execute the search
         search = self._search(
             "search",
             identity,
             params,
             es_preference,
-            permission_action="read_event",
+            permission_action="unused",
```
lnielsen, line 143: Invenio-Records-Resources should be changed to allow not specifying a permission action.

invenio_requests/services/permissions.py
```diff
@@ -11,12 +11,83 @@
...
-    # For search, recall that _what_ identities can see is defined by `can_read`
-    # All other `can_` are taken care by the request types
+    # Ability in general to create requests (not which request you can create)
+    can_create = [AuthenticatedUser(), SystemProcess()]
+    # Just about ability to perform a search (not what requests you can access)
     can_search = [AuthenticatedUser(), SystemProcess()]
+
+    # Read/update/delete action deals with requests in **multiple states**, and
+    # thus must take the request status into account.
+    can_read = [
```
lnielsen, line 34: In the end, I found this way to be the best trade-off between performance and readability.

1) Searches over record that a user can access must take the request status into account
2) By having only two status generators we essentially are able to generate queries like below:

```
(
   status:(created) 
   AND grants:(created_by.1)
)
OR
(
   status:(submitted, deleted, cancelled, expired, accepted, declined) 
   AND grants:(created_by.id.1, receiver.community.a.manager, receiver.community.a.owner)
)
```

invenio_requests/services/generators.py
```diff
@@ -0,0 +1,98 @@
...
+    def needs(self, request=None, **kwargs):
+        """Needs for the given entity reference."""
+        entity = getattr(request, self.entity_field)
+        return request.type.entity_needs(entity)
+
+    def query_filter(self, identity=None, **kwargs):
+        """Query filters for the current identity."""
+        grants = []
+        for need in identity.provides:
+            grants.append(EntityGrant(self.entity_field, need).token)
```
lnielsen, line 69: Grant tokens take the form:

``created_by.id.1`` or ``receiver.community.9ccd04c5-d6d9-4c7d-ae25-e54a48ea66ad.manager``

E.g. a need like ``UserNeed(1)`` is just a named tuple ``('id', 1)`` and the need ``CommunityRoleNeed('9ccd04c5-d6d9-4c7d-ae25-e54a48ea66ad', 'manager') is a named tuple ``('community', '9ccd04c5-d6d9-4c7d-ae25-e54a48ea66ad', 'manager').

This a grant token is just a prefix + the tuple elements concatenated

- ``created_by`` + ``UserNeed(1)`` -> 
- ``created_by`` + ``('id', 1)`` -> 
- ``created_by.id.1``

PR #160, opened 2022-03-01
permissions: disable requests when communities not enabled

ppanero:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1315

**Communities OFF**

Screenshot 2022-03-01 at 13 20 44

Screenshot 2022-03-01 at 13 20 27

**Communities ON**

Screenshot 2022-03-01 at 13 13 47

Screenshot 2022-03-01 at 13 14 50

(missing tests)

### DIFF
--- a/tests/resources/requests/test_requests_resources.py
+++ b/tests/resources/requests/test_requests_resources.py
@@ -197,3 +197,18 @@ def test_simple_request_flow(app, client_logged_as, headers, example_request):
         }
     )
     assert_api_response(response, 200, expected_data)
+
+
+def test_disabled_endpoints(app, client_logged_as, headers, example_request):
+    app.config["COMMUNITIES_ENABLED"] = False
+    client = client_logged_as("user1@example.org")
+    id_ = str(example_request.id)
+
+    response = client.get(f"/requests/{id_}", headers=headers)
+    assert response.status_code == 403
+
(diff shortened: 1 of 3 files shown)


tests/resources/requests/test_requests_resources.py
```diff
@@ -197,3 +197,24 @@ def test_simple_request_flow(app, client_logged_as, headers, example_request):
...
+    id_ = str(example_request.id)
+
+    response = client.get(f"/requests/{id_}", headers=headers)
+    assert response.status_code == 403
+
+    response = client.post(
+        f"https://127.0.0.1:5000/api/requests/{id_}/actions/submit",
+        headers=headers
+    )
+    assert response.status_code == 404  # id not found
```
ppanero, line 214: The request gets searched for before permissions are checked:

```
        request = self.record_cls.get_record(id_)
        action_obj = RequestActions.get_action(request, action)

        # check permissions
        permission_name = f"action_{action}"
        self.require_permission(identity, permission_name, request=request)
```

ppanero:
@lnielsen as we discussed, maybe is preferrable to have it done by not having the endpoint directly but it not feasible unless we start playing with blueprint registration (i.e. we need to make sure config is loaded before, to have the env var. we could do it in the views.py... but is hackish).

Issue #129, opened 2022-02-08
Request Details: Pagination for conversation timeline


NRodriguezcuellar:
Add pagination to the conversation section of the Request details page

PR #163, opened 2022-03-07
Request Details Timeline: pagination

NRodriguezcuellar:
:heart: Thank you for your contribution!

### Description
Added pagination to the request details conversation 

closes https://github.com/inveniosoftware/invenio-requests/issues/129

Screenshot from 2022-03-07 13-09-02

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/TimelineFeed.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/TimelineFeed.js
@@ -6,6 +6,7 @@
 
 import Error from "../components/Error";
 import Loader from "../components/Loader";
+import { Pagination } from "../components/Pagination";
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import Overridable from "react-overridable";
@@ -39,7 +40,7 @@ class TimelineFeed extends Component {
   };
 
   render() {
-    const { timeline, loading, error } = this.props;
+    const { timeline, loading, error, setPage, size, page } = this.props;
     const { modalOpen, modalAction } = this.state;
 
     return (
@@ -47,7 +48,7 @@ class TimelineFeed extends Component {
         <Error error={error}>
           <Overridable id="TimelineFeed.layout" {...this.props}>
             <Container>
-              <Segment>
+              <Segment className="borderless shadowless">
                 <Feed>
                   {timeline.hits?.hits.map((comment) => (
                     <TimelineEventWithState
@@ -58,6 +59,15 @@ class TimelineFeed extends Component {
                   ))}
                 </Feed>
                 <Divider />
+                <Container textAlign="center">
+                  <Pagination
+                    page={page}
+                    size={size}
+                    setPage={setPage}
+                    totalLength={timeline.hits?.total}
+                  />
+                </Container>
+                <Divider hidden />
                 <TimelineCommentEditor />
                 <DeleteConfirmationModal
                   open={modalOpen}
@@ -80,6 +90,9 @@ TimelineFeed.propTypes = {
   timeline: PropTypes.object,
   error: PropTypes.object,
   isSubmitting: PropTypes.bool,
+  setPage: PropTypes.func.isRequired,
+  page: PropTypes.number,
+  size: PropTypes.number,
 };
 
 export default Overridable.component("TimelineFeed", TimelineFeed);
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
@@ -8,6 +8,7 @@ export const IS_LOADING = "timeline/IS_LOADING";
 export const SUCCESS = "timeline/SUCCESS";
 export const HAS_ERROR = "timeline/HAS_ERROR";
 export const IS_REFRESHING = "timeline/REFRESHING";
+export const CHANGE_PAGE = "timeline/CHANGE_PAGE";
 
 class intervalManager {
   static IntervalId = undefined;
@@ -17,8 +18,11 @@ class intervalManager {
   }
 }
 
-export const fetchTimeline = (loadingState = true, chosenParams) => {
+export const fetchTimeline = (loadingState = true) => {
   return async (dispatch, getState, config) => {
+    const state = getState();
+    const { size, page } = state.timeline;
+
     if (loadingState) {
       dispatch({
         type: IS_LOADING,
@@ -29,7 +33,11 @@ export const fetchTimeline = (loadingState = true, chosenParams) => {
     });
 
     try {
-      const response = await config.requestsApi.getTimeline(chosenParams);
+      const response = await config.requestsApi.getTimeline({
+        size: size,
+        page: page,
+        sort: 'oldest',
+      });
       dispatch({
         type: SUCCESS,
         payload: response.data,
@@ -40,6 +48,17 @@ export const fetchTimeline = (loadingState = true, chosenParams) => {
         payload: error,
       });
     }
+  };
+};
+
+export const setPage = (page) => {
+  return async (dispatch, getState, config) => {
+    dispatch({
+      type: CHANGE_PAGE,
+      payload: page,
+    });
+
+    dispatch(fetchTimeline());
   };
 };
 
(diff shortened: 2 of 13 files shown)


invenio_requests/assets/semantic-ui/js/invenio_requests/api/api.js
```diff
@@ -6,5 +6,6 @@
 export {
   RequestEventsLinksExtractor,
   InvenioRequestEventsApi,
+  QueryBuilder,
```
Pineirin, line 9: What's this for?

invenio_requests/assets/semantic-ui/js/invenio_requests/api/index.js
```diff
@@ -7,5 +7,6 @@ export {
   RequestLinksExtractor,
   InvenioRequestEventsApi,
   InvenioRequestsAPI,
+  QueryBuilder,
```
Pineirin, line 10: Same here.

Pineirin (approved):

invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEditor/state/actions.js
```diff
@@ -32,11 +38,27 @@ export const submitComment = (content, format) => {
...
-      dispatch({
+      const queryParams = new QueryBuilder(timelineState.queryParams);
+      const currentCommentsLength = timelineState.data.hits.hits.length;
+      const shouldGoToNextPage = currentCommentsLength + 1 > queryParams.size;
+
+      if (shouldGoToNextPage) {
+        const nextPage = queryParams.page + 1;
+        const newQueryParams = queryParams.setPage(nextPage).get();
+
+        dispatch({ type: SETTING_PARAMS, payload: newQueryParams });
```
ntarocco, line 49: As discussed, hide internal implementations to the outside world and expose clear APIs, eg.:
```suggestion
        dispatch({ type: CHANGE_PAGE, payload: {currentPage: 3} });
```

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/TimelineFeed.js
```diff
@@ -39,15 +40,16 @@ class TimelineFeed extends Component {
...
     return (
       <Loader isLoading={loading}>
         <Error error={error}>
           <Overridable id="TimelineFeed.layout" {...this.props}>
             <Container>
-              <Segment>
+              <Segment id="request-details-feed-segment">
                 <Feed>
                   {timeline.hits?.hits.map((comment) => (
                     <TimelineEventWithState
```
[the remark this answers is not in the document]
NRodriguezcuellar: we have a TimelineEvent component that only accepts props and then also a TimelineEventWithState which holds a local state and connects to the redux actions.

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
```diff
@@ -43,6 +51,19 @@ export const fetchTimeline = (loadingState = true, chosenParams) => {
   };
 };
 
+export const setPage = (page, refresh = true) => {
```
ntarocco, line 54: Looks like `refresh` param is never called?
```suggestion
export const setPage = (page) => {
```

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
```diff
@@ -62,12 +83,17 @@ const timelineReload = (dispatch, getState, config) => {
...
-    const intervalId = setInterval(
-      () => timelineReload(dispatch, getState, config),
-      config.refreshIntervalMs
-    );
-    intervalManager.setIntervalId(intervalId);
+    dispatch(fetchTimeline(true));
+
+    const intervalAlreadySet = intervalManager.intervalId;
+
+    if (!intervalAlreadySet) {
```
ntarocco, line 90: do you still need this `if`? It is just to be safe?

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/reducer.js
```diff
@@ -4,13 +4,22 @@
...
+  IS_REFRESHING,
+  SETTING_PARAMS,
+} from "./actions";
 
 export const initialState = {
   loading: false,
   refreshing: false,
   data: {},
   error: null,
+  size: 10,
```
kpsherva, line 20: this is quite hidden as a configuration, hard to find in case of backend changes. Did you check if it is possible to set and retrieve this config from the backend ?

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
NRodriguezcuellar: it is not really needed, it is just to be safe at this point.

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/reducer.js
NRodriguezcuellar: that is a fair point, but then setting in the backend i am not sure if that makes more sense than setting it in the appConfig (frontend). What do you think?

kpsherva: I think it makes more sense since we pass all the search configs from the backend in all the other cases, so I would expect the same here, maybe we can hear other opinions also @ntarocco @Pineirin @zzacharo @jrcastro2

ntarocco: Ye, the `size` yes it can be injected and configured from the backend. However, the `sort` must be hardcoded because otherwise the appending of the new comment might fail, right @NRodriguezcuellar?
If that's the case, I would hardcoded it in the URL before calling it and not make it changeable in Redux.

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
```diff
@@ -43,6 +51,17 @@ export const fetchTimeline = (loadingState = true, chosenParams) => {
...
 };
 
+export const setPage = (page) => {
+  return async (dispatch, getState, config) => {
+    dispatch({
+      type: CHANGE_PAGE,
+      payload: page,
+    });
+
+    dispatch(fetchTimeline());
```
kpsherva, line 61: shouldn't fetchTimeline have page as argument ? I find it a bit hidden to retrieve it from the state inside of the fetchTimeline function. Was there any reason to implement it like that?
I think if you set page aa param then you can keep it locally as component state, not in redux, it will be probably simpler to manage

invenio_requests/assets/semantic-ui/less/invenio_requests/theme.less
```diff
@@ -1,3 +1,8 @@
 .ui.segment.loader-container {
   height: 10em;
 }
+
+#request-details-feed-segment {
```
kpsherva, line 5: theme.less will be deleted, please move it to site.overrides as:
```suggestion
.shadowless{
box-shadow: none !imporant;
}
```

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/TimelineFeed.js
```diff
@@ -39,15 +40,15 @@ class TimelineFeed extends Component {
...
+    const { timeline, loading, error, setPage, size, page } = this.props;
     const { modalOpen, modalAction } = this.state;
 
     return (
       <Loader isLoading={loading}>
         <Error error={error}>
           <Overridable id="TimelineFeed.layout" {...this.props}>
             <Container>
-              <Segment>
+              <Segment id="request-details-feed-segment">
```
kpsherva, line 51:
```suggestion
              <Segment className="borderless shadowless">
```
see below in CSS for the full explanation

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
NRodriguezcuellar: It should be in the universal state because it gets used in multiple places, like in the redux action to calculate if the page needs to be skipped. Also it would need to be passed down in every place that the fetch timeline function is being used, which is error prone in my opinion.

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/TimelineFeed.js
NRodriguezcuellar: sure :+1:

ntarocco (approved):

ntarocco (approved):

zzacharo (approved):

Issue #156, opened 2022-02-24
requests ui: commenting permissions


kpsherva:
Currently we display the edit/delete actions for each comment, however the user should not be able to perform any actions on a comment which was not written originally by the user.
The actions should not be visible if the user has no permissions to perform edit/delete of a specific comment. 

Solution 1:
A user should only be available to edit or delete a request (in this case a comment) it it has the permissions required for those actions, which, as defined in the configuration https://github.com/inveniosoftware/invenio-requests/blob/2f3a281c7f9718d0df30bae9db4a84a94b6c926d/invenio_requests/services/permissions.py#L81-L88, require, in both cases, that the request was created by the user.

Getting this permissions in the _frontend_ can be, as far as I have tested, a bit problematic, and, in my opinion, the best solution would be to create  and API endpoint to obtain the permissions given an object and a record. This solution has been discussed and may be quite overkill and complex for such a small feature. 

Solution 2:
A way simpler and even cleaner solution would be to compare the created_by field of the comment with the id of the logged user, showing the edit/delete dropdown if both match.

To discuss:
- should a community manager/admin be able to edit or delete someone's comment?

ntarocco:
As discussed, investigate a 3rd solution to expose the permissions of the current logged in user for each event, as done for the members.

PR #172, opened 2022-03-30
Permissions: return permissions in request event calls

Pineirin:
Display edit/delete dropdown only for authorized users.
closes: inveniosoftware/invenio-requests#156

View of the user admin@test.ch
Screenshot from 2022-03-29 15-56-23

View of the user pablo@gmail.com
Screenshot from 2022-03-29 15-56-30

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvent/TimelineEvent.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvent/TimelineEvent.js
@@ -44,6 +44,9 @@ class TimelineEvent extends Component {
     const commentHasBeenDeleted = !event.payload;
     const commentCanBeDeleted = event.payload;
 
+    const canDelete = event.permissions.can_delete_comment;
+    const canUpdate = event.permissions.can_update_comment;
+
     return (
       <Overridable id="TimelineEvent.layout" event={event}>
         <Feed.Event>
@@ -68,15 +71,19 @@ class TimelineEvent extends Component {
                   </Feed.Summary>
                 </Grid.Column>
                 <Grid.Column width={1}>
-                  {commentCanBeDeleted && (
+                  {commentCanBeDeleted && (canDelete || canUpdate) && (
                     <Dropdown icon="ellipsis horizontal">
                       <Dropdown.Menu>
-                        <Dropdown.Item onClick={() => toggleEditMode()}>
-                          {i18next.t("Edit")}
-                        </Dropdown.Item>
-                        <Dropdown.Item onClick={() => deleteComment()}>
-                          {i18next.t("Delete")}
-                        </Dropdown.Item>
+                        {canUpdate && (
+                          <Dropdown.Item onClick={() => toggleEditMode()}>
+                            {i18next.t("Edit")}
+                          </Dropdown.Item>
+                        )}
+                        {canDelete && (
+                          <Dropdown.Item onClick={() => deleteComment()}>
+                            {i18next.t("Delete")}
+                          </Dropdown.Item>
+                        )}
                       </Dropdown.Menu>
                     </Dropdown>
                   )}
--- a/invenio_requests/services/events/service.py
+++ b/invenio_requests/services/events/service.py
@@ -11,15 +11,21 @@
 
 from elasticsearch_dsl import Q
 from invenio_access.permissions import system_process
-from invenio_records_resources.services import RecordService
+from invenio_records_resources.services import RecordService, ServiceSchemaWrapper
 from invenio_records_resources.services.base.links import LinksTemplate
 from invenio_records_resources.services.uow import RecordCommitOp, unit_of_work
 
 from ...records.api import RequestEventType
+from ..schemas import RequestEventDumpSchema
 
 
 class RequestEventsService(RecordService):
     """Request Events service."""
+
+    @property
+    def event_dump_schema(self):
+        """Schema for creation."""
+        return ServiceSchemaWrapper(self, schema=RequestEventDumpSchema)
 
     @unit_of_work()
     def create(self, identity, request_id, data, uow=None):
@@ -56,6 +62,7 @@ class RequestEventsService(RecordService):
             identity,
             event,
             links_tpl=self.links_item_tpl,
+            schema=self.event_dump_schema
         )
 
     def read(self, identity, id_):
@@ -70,6 +77,7 @@ class RequestEventsService(RecordService):
             identity,
             record,
             links_tpl=self.links_item_tpl,
+            schema=self.event_dump_schema
         )
 
     @unit_of_work()
@@ -101,6 +109,7 @@ class RequestEventsService(RecordService):
             identity,
             event,
             links_tpl=self.links_item_tpl,
+            schema=self.event_dump_schema
         )
 
     @unit_of_work()
@@ -156,6 +165,7 @@ class RequestEventsService(RecordService):
                 context={"request_id": request_id, "args": params},
             ),
             links_item_tpl=self.links_item_tpl,
+            schema=self.event_dump_schema
         )
 
     # Utilities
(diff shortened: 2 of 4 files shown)


invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
...
         """Return key of type_schemas to use given object to dump."""
         return obj.type
+
+
+class Permissions():
+    """Add permissions to the dump data."""
+
+    permissions = fields.Method("get_permissions")
+
+    def is_self(self, obj):
```
kpsherva, line 130: I think this name can be interpreted in too many ways, could you change it into something more explanatory, like `is_comment_creator` `in_event_creator` ?

setup.cfg
```diff
@@ -35,6 +35,10 @@ tests =
     invenio-db[postgresql,mysql,versioning]>=1.0.13,<2.0
     pytest-invenio>=1.4.3
     sphinx>=4.2.0,<5
+    Flask<=2.0.3
```
kpsherva, line 38: why did you need flask to be installed in the module ?

tests/services/events/test_request_events_service.py
```diff
@@ -55,6 +55,8 @@ def test_simple_flow(
     # Update it
     data = read_item.to_dict()  # should be equivalent to events_service_data
     data["payload"]["content"] = "An edited comment"
+    # Remove permissions from data as they wont be persisted.
+    data.pop("permissions")
```
kpsherva, line 59: not sure why tou pop the permissions?

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
     def get_obj_type(self, obj):
         """Return key of type_schemas to use given object to dump."""
         return obj.type
+
+
+class Permissions():
```
kpsherva, line 125: I think the name should follow the pattern of other schema classes fe `EventPermissionShema`

invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvent/TimelineEvent.js
```diff
@@ -68,15 +71,19 @@ class TimelineEvent extends Component {
...
-                        </Dropdown.Item>
-                        <Dropdown.Item onClick={() => deleteComment()}>
-                          {i18next.t("Delete")}
-                        </Dropdown.Item>
+                        {canUpdate && (
+                          <Dropdown.Item onClick={() => toggleEditMode()}>
+                            {i18next.t("Edit")}
+                          </Dropdown.Item>
+                        )}
+                        {canUpdate && (
```
kpsherva, line 82:
```suggestion
                        {canDelete && (
```

kpsherva (commented):
just because the solution is changing the expected payload of the event, was this discussed with anyone, or taken into the account in the RFCs?

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
...
         return obj.type
+
+
+class EventPermissionShema():
+    """Add permissions to the dump data."""
+
+    permissions = fields.Method("get_permissions")
+
+    def in_event_creator(self, obj):
+        """True if the creator of the record is the identified user."""
```
ntarocco, line 131:
```suggestion
        """True if the creator of the record is the current_user."""
```

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
...
         """Return key of type_schemas to use given object to dump."""
         return obj.type
+
+
+class EventPermissionShema():
+    """Add permissions to the dump data."""
+
+    permissions = fields.Method("get_permissions")
+
+    def in_event_creator(self, obj):
```
ntarocco, line 130:
```suggestion
    def is_event_creator(self, obj):
```

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
...
+        """True if the creator of the record is the identified user."""
+        if "is_self" not in self.context:
+            current_identity = self.context["identity"]
+            user_id = obj.created_by._resolve().get_id()
```
ntarocco, line 134: I am not sure about this code, why the `_resolve()`?
What about:
```
def is_event_creator(self, obj):
    """...."""
    current_identity = self.context["identity"]
    user_id = obj.created_by.get('user', '')
    return current_identity.id is not None and str(user_id) == str(current_identity.id)
```

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
...
+            self.context["is_self"] = (
+                user_id is not None
+                and current_identity.id is not None
+                and str(user_id) == str(current_identity.id)
+            )
+        return self.context["is_self"]
+
+    def get_permissions(self, obj):
+        """Get certain permissions of the user with the current hit."""
+        return {
```
ntarocco, line 144:
```suggestion
        is_creator = self.in_event_creator(obj)
        return {
            can_update_comment: is_creator,
            can_delete_commet: is_creator,
```

invenio_requests/services/events/service.py
```diff
@@ -11,16 +11,22 @@
...
 
 from ...records.api import RequestEventType
+from ..schemas import RequestEventDumpSchema
 
 
 class RequestEventsService(RecordService):
     """Request Events service."""
 
+    @property
+    def member_dump_schema(self):
```
ntarocco, line 26: rename?

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
...
+
+class CommentDumpSchema(CommentSchema, EventPermissionShema):
+    """Comment dump schema."""
+
+
+class NoExtrasDumpSchema(NoExtrasSchema, EventPermissionShema):
+    """Comment dump schema."""
+
+
+class RequestEventDumpSchema(BaseRecordSchema, OneOfSchema):
```
ntarocco, line 158: I am not sure this duplication of schemas is needed, let's discuss IRL.

invenio_requests/services/schemas.py
```diff
@@ -120,3 +120,54 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
     def get_obj_type(self, obj):
         """Return key of type_schemas to use given object to dump."""
         return obj.type
+
+
+class EventPermissionShema():
```
ntarocco, line 125: Implementation is nice, good work, I am just thinking that `can_update_comment` or `can_delete_comment` are permissions that are configurable, while here you are hardcoding that the current logged in user must be the creator.

jrcastro2 (approved):
LGTM! :rocket:

tests/services/events/test_request_events_service.py
lnielsen: Is it because you cannot send ``permissions``  key back? If, so then it's likely a bug that should be fixed in the schema.......

invenio_requests/services/schemas.py
```diff
@@ -120,3 +123,47 @@ class RequestEventSchema(BaseRecordSchema, OneOfSchema):
     def get_obj_type(self, obj):
         """Return key of type_schemas to use given object to dump."""
         return obj.type
+
+
+class EventPermissionShema(Schema):
```
ntarocco, line 128: It looks to me very Comment specific, probably rename to:
```suggestion
class CommentPermissionShema(Schema):
```

ntarocco (approved):

PR #179, opened 2022-04-12
refactor request event type

### LINKED ISSUE #181: Add non-comments events to events
Events that are not comments are not returned by the backend.

ntarocco:
closes #181

### DIFF
--- a/invenio_requests/__init__.py
+++ b/invenio_requests/__init__.py
@@ -11,8 +11,9 @@
 from .customizations import RequestAction
 from .ext import InvenioRequests
 from .proxies import (
+    current_event_type_registry,
     current_events_service,
-    current_registry,
+    current_request_type_registry,
     current_requests,
     current_requests_resource,
     current_requests_service,
@@ -22,8 +23,9 @@ __version__ = "0.3.9"
 
 __all__ = (
     "__version__",
+    "current_event_type_registry",
     "current_events_service",
-    "current_registry",
+    "current_request_type_registry",
     "current_requests_resource",
     "current_requests_service",
     "current_requests",
--- a/invenio_requests/customizations/actions.py
+++ b/invenio_requests/customizations/actions.py
@@ -7,8 +7,9 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Base class for customizable actions on requests."""
-
 from ..errors import NoSuchActionError
+from ..proxies import current_events_service
+from .event_types import LogEventType
 
 
 class RequestAction:
@@ -21,7 +22,10 @@ class RequestAction:
     """Status after execution of the action."""
 
     event_type = None
-    """Defines an event type which will be logged if defined."""
+    """Defines an event which will be logged if defined."""
+
+    log_event = True
+    """Define if the action should be logged through a `LogEventType`."""
 
     def __init__(self, request):
         """Constructor."""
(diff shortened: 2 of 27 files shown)


ntarocco:
@lnielsen did you mean something like this? I am not sure I correctly read your mind.
The advantage is extensibility, but there still the need to refer to/harcode `Comments` request event type around.
To discuss IRL.

tests/resources/events/test_request_events_resources.py
```diff
@@ -129,16 +130,19 @@ def test_simple_comment_flow(
...
     expected_json_3 = {
         "created_by": {"user": "2"},
-        "id": comment_id,
-        "links": {
-            "self": f"https://127.0.0.1:5000/api/requests/{request_id}/comments/{comment_id}",  # noqa
+        "revision_id": 1,
+        "payload": {
+            "content": "deleted a comment",
+            "format": "html",
+            "event": "COMMENT_DELETED"
```
zzacharo, line 137: @lnielsen here we can see that the default search returns also the deleted comments. Shall we handle the filtering on the frontend or do we want to add a filter in the service method so we can filter them out or return them?

invenio_requests/services/requests/service.py
```diff
@@ -213,24 +214,30 @@ def execute_action(self, identity, id_, action, data=None, uow=None):
...
-        # Because the action may change the request's status, this has to be done
-        # before the action is executed
+        # Because the action may change the request's status, this has to be
+        # done before the action is executed
+
         event_type = action_obj.event_type
         if event_type is not None:
-            current_events_service.create(
-                identity, request.id, {"type": event_type}, uow=uow
+            _data = dict(
```
lnielsen, line 222: This part

invenio_requests/customizations/actions.py
```diff
@@ -100,6 +100,7 @@ class DeleteAction(RequestAction):
 
     status_from = ['created']
     status_to = 'deleted'
+    event_type = LogEventType(payload=dict(event="DELETED"))
```
lnielsen, line 103: I would match the name here, up with the names in ``RequestType.available_statuses`` - i.e. lowercase instead of upper case.

Also, I think I would move the logic to the ``execute()`` method of the action instead of doing it as part of the service. 

```python
class RequestAction: 
    log_event = True

    def execute(self, identity, uow):
        self.request.status = self.status_to
        if self.log_event:
            event = LogEventType(event=self.status_to). # <---- using self.status_to
            current_events_service.create(identity, request.id, _data, event_type, uow=uow)


# ...


class DeleteAction
     status_from = ['created']
     status_to = 'deleted'
     log_event = False
```

invenio_requests/__init__.py
```diff
@@ -12,7 +12,7 @@
 from .ext import InvenioRequests
 from .proxies import (
     current_events_service,
-    current_registry,
+    current_request_type_registry,
```
lnielsen, line 15: Comment: For consistency I would also export ``current_event_type_registry``.

tests/resources/events/test_request_events_resources.py
lnielsen: I think I discussed this with @ntarocco. We decided that the best is that we simply delete the full comment, and log new event, that user x deleted a comment. So we don't keep the previous comment around with empty value, but instead write a new event

zzacharo: sorry that's what I meant.... now when you will search, you will reveive the "user deleted a comment" events too so I was wondering if that would be something that it was useful to show to the UI and if not I was trying to find a solution to filter them out on demand... but from what you are saying the implementation is there!

tests/records/systemfields/test_event_type_systemfield.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: You can remove this file.

lnielsen (approved):

PR #198, opened 2022-04-27
timeline: connect events backend

### LINKED ISSUE #183: [BLOCKED] ui: integrate events backend
Integrate the new events when https://github.com/inveniosoftware/invenio-requests/issues/181 is implemented and remove mocked data

zzacharo:
closes #183


invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvents/timelineActionEvents.js
```diff
@@ -6,28 +6,51 @@
...
+    event={event}
+    eventContent={i18next.t("cancelled this request")}
+    iconColor="red"
+  />
+);
+
+export const TimelineUnknownEvent = ({ event }) => (
+  <TimelineActionEvent
+    iconName="close"
+    iconColor={"red"}
```
kpsherva, line 51: nit: formatting

invenio_requests/customizations/event_types.py
```diff
@@ -115,23 +115,37 @@ class LogEventType(EventType):
...
-            validate=validate.Length(min=1)
+    def payload_schema():
+        """Return payload schema as a dictionary"""
+        from invenio_requests.records.api import RequestEventFormat
+        return dict(
+            event=fields.String(validate=validate.Length(min=1)),
+            content=utils_fields.SanitizedHTML(
+                validate=validate.Length(min=1)
+            ),
+            format=fields.Str(
```
kpsherva, line 126: just for me to understand how this works, why was the format moved from EventType to LogEventType? (we can discuss in a call, not a blocker)

zzacharo: So, I wanted to implement the feature of passing the payload schema in the inherited class but because of circular imports I had to do it in the EventType class...That said, it was bugging me because I was forcing to have a format field to the passed payload so I changed the implementation to support a callable as `payload_schema` so I can have a local import

kpsherva: do you think it makes sense to add a comment in the code to explain this ? or a bit overkill ?

kpsherva (approved):

Issue #204, opened 2022-05-04
Fix UI bugs in request details page


jrcastro2:
Package version (if known): v9.0.0.dev14

## Describe the bug
Fix the various bugs in the UI for requests details page

## Steps to Reproduce
1. Go to details page of any request
2. See long text being cut
3. Try to edit a comment and see the overflow
4. Actions are missing the first capital letter
5. The cancel modal contains the word cancel twice. Maybe a solution could be to use "Cancel" and "Confirm" for all the actions.

## Screenshots (if applicable)
![image](https://user-images.githubusercontent.com/15194802/166672998-e5a42893-d55a-4adf-9599-37fc098b7bc6.png)

![image](https://user-images.githubusercontent.com/15194802/166673010-c44a09cb-79b1-416d-a15a-333d377cb65e.png)

kpsherva:
the second screenshot probably needs also request actions translations to be implemented, I would suggest opening a separate issue for that?

jrcastro2:
still issue on comments 
![image](https://user-images.githubusercontent.com/15194802/169524260-cc439fc0-56d5-40a7-af8c-384440d64d30.png)

PR #208, opened 2022-05-09
request details: fix layout bugs and adjust for mobile and tablet

jennur:
Closes https://github.com/inveniosoftware/invenio-requests/issues/204

- Reduced the amount of nested `.ui.container`s, as these are all getting horizontal margins when resizing to tablet and mobile (=> the content becomes very narrow on smaller screens).
- Changed the column widths for mobile/tablet
- Added correct level headings

## Related PR
https://github.com/inveniosoftware/invenio-app-rdm/pull/1523

## Screenshots
127 0 0 1_5000_communities_comp_requests_7dd0f125-c3ed-4d15-9b9d-9fa50b0d0c21 (1)
127 0 0 1_5000_communities_comp_requests_7dd0f125-c3ed-4d15-9b9d-9fa50b0d0c21 (2)
127 0 0 1_5000_communities_comp_requests_7dd0f125-c3ed-4d15-9b9d-9fa50b0d0c21

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEditor/TimelineCommentEditor.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEditor/TimelineCommentEditor.js
@@ -19,14 +19,14 @@ const TimelineCommentEditor = ({
   submitComment,
 }) => {
   return (
-    <Container className="timeline-comment-editor-container">
+    <div className="timeline-comment-editor-container">
       {error && <Message negative>{error}</Message>}
       <FormattedInputEditor
         data={commentContent}
         onChange={(event, editor) => setCommentContent(editor.getData())}
         minHeight="7rem"
       />
-      <Container className="mt-10" textAlign="right">
+      <div className="text-align-right rel-mt-1">
         <SaveButton
           icon="send"
           size="medium"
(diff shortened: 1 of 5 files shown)


invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEditor/TimelineCommentEditor.js
```diff
@@ -19,23 +19,23 @@ const TimelineCommentEditor = ({
   submitComment,
 }) => {
   return (
-    <Container className="timeline-comment-editor-container">
+    <div className="timeline-comment-editor-container">
```
kpsherva, line 22: Why is it changed back to div? we should aim to use the semantic ui react components instead of pure html in react

jennur: Because the `<Container />` comes with default styling which is not always wanted or necessary. It will create a lot of cases where we will have to override spacing if we need to avoid using pure html. In this case we had 2 or 3 nested Containers where each container has a horizontal margin of `1rem` on mobile, and this would have to be overridden as the content got super narrow.

jennur: What is the reason we want to avoid pure html in React?

jennur:
The white space on the mobile screenshot is something that happens in the Chrome screenshot tool. I double checked it, and nothing seem to overflow there :)

ntarocco (approved):

PR #235, opened 2022-05-23
metadata: change title of creator

NRodriguezcuellar:
change proposed in https://github.com/inveniosoftware/invenio-app-rdm/issues/1596


invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvents/TimelineCommentEvent.js
```diff
@@ -67,13 +67,13 @@ class TimelineCommentEvent extends Component {
     if (isUser) {
       userAvatar = (
         <RequestsFeed.Avatar
-          src={expandedCreatedBy.links.avatar}
+          src={expandedCreatedBy?.links?.avatar}
```
ntarocco, line 70: such changes should not be needed. Why did you add them?
If the avatar or full_name are not set, there is a problem...

NRodriguezcuellar: Because it breaks when i use an admin account created with an `users.yml` file as stated in https://inveniordm.docs.cern.ch/customize/vocabularies/users/

ntarocco: Then the solution is to add the user profile to the accounts created via the vocabulary... Probably somewhere [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/users.py#L56)... or for another pR.
The code must fail if the profile is not set, so we can detect if there is an error (and not silently failing). The avatar, when not set, should fallback to a default icon.

NRodriguezcuellar: I did not consider that an `error` could be wanted, i think the profile should be added indeed. Create a new issue for that?

NRodriguezcuellar: https://github.com/inveniosoftware/invenio-app-rdm/issues/1658

ntarocco (approved):

PR #243, opened 2022-06-07
requestactions: add disabled when loading

jrcastro2:
* Add return of action promise
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1215

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/request/actions/RequestAction.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/request/actions/RequestAction.js
@@ -30,7 +30,7 @@ export class RequestAction extends Component {
     const { performAction } = this.context;
     const { action } = this.props;
     const { actionComment } = this.state;
-    performAction(action, actionComment);
+    return performAction(action, actionComment);
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


Pineirin (approved):

invenio_requests/assets/semantic-ui/js/invenio_requests/request/actions/RequestAction.js
```diff
@@ -30,7 +30,7 @@ export class RequestAction extends Component {
     const { performAction } = this.context;
     const { action } = this.props;
     const { actionComment } = this.state;
-    performAction(action, actionComment);
+    return performAction(action, actionComment);
```
kpsherva, line 33: I think there is no need to return this function call

jrcastro2: If I am not mistaken it is needed mainly because we want to wait to get the response back from the backend (like [here](https://github.com/inveniosoftware/invenio-app-rdm/pull/1700/files#diff-90ca62aa76c33b515af46cf068a5b55ad042577f7cfac924e5dd5c77ce664319R505)) to be able to re-trigger the search again, after the action was indexed in ES, to do so we need to wait till this action was performed. If we don't return the promise here and we would wait for this method to finish it will finish (most-likely) before the action was indexed in ES (as we don't wait for the promise to be resolved), but if we want to make sure  that the action was indexed I do think the return is necessary.

PR #248, opened 2022-06-14
1668 dont override request

fenekku:
- templates: don't override global request in Jinja templates
- extras_require: remove elasticsearch6
- close https://github.com/inveniosoftware/invenio-app-rdm/issues/1668

### DIFF
--- a/invenio_requests/templates/semantic-ui/invenio_requests/details/index.html
+++ b/invenio_requests/templates/semantic-ui/invenio_requests/details/index.html
@@ -26,7 +26,7 @@
           <div class="sixteen wide mobile ten wide tablet eight wide computer column">
             <h2 class="ui header">
               {% block request_header_title %}
-                {{ rdm_request.title }}
+                {{ invenio_request.title }}
               {% endblock request_header_title %}
             </h2>
           </div>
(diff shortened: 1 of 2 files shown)


invenio_requests/templates/semantic-ui/invenio_requests/details/index.html
```diff
@@ -26,7 +26,7 @@
           <div class="sixteen wide mobile ten wide tablet eight wide computer column">
             <h2 class="ui header">
               {% block request_header_title %}
-                {{ request.title }}
+                {{ rdm_request.title }}
```
kpsherva, line 29: I think it is a bit strange to have rdm mentioned here since this module does not depend on RDM, could you give us a bit more context why you need this change ? I am not sure if I understand it properly

fenekku: The original problem is that this overrides `request` provided by Jinja: https://flask.palletsprojects.com/en/2.1.x/templating/#standard-context . In https://github.com/inveniosoftware/invenio-app-rdm/pull/1739 I've used `rdm_request` in the templates, so I am using it here too :shrug: . As long as the same name is used everywhere, it works. Another suggestion was `invenio_request` but maybe that's too close to the module name...

kpsherva: ah interesting bug :) thanks for spotting this, there is totally namespacing issue. I agree `invenio_request` can also be problematic, but I cannot think about other more suitable name (maybe it is OK, since the module name is in plural form?)

kpsherva (approved):

kpsherva (approved):

Issue #251, opened 2022-06-15
comment deletion not being handled properly


NRodriguezcuellar:
![Screenshot from 2022-06-15 14-08-03](https://user-images.githubusercontent.com/55200060/173823740-0aeeb496-65fa-495f-b81a-d0d28361d4df.png)

steps to reproduce
- delete a comment
- It will show that the comment has been deleted
![Screenshot from 2022-06-15 14-08-30](https://user-images.githubusercontent.com/55200060/173823650-0aa2a105-3587-41cb-a914-910877c5662c.png)
- after some time it will refresh from the backend and then the  comment will removed and a new event is pushed to the back of the comments
![Screenshot from 2022-06-15 14-08-13](https://user-images.githubusercontent.com/55200060/173823871-61e62567-609b-4cb1-98e6-fcd5ef85b28b.png)

this is very confusing for the user, right now the backend and frontend are out of sync in terms of what they expect to happen when comments are deleted. Before, the behaviour was that they were not completely removed, just the payload was removed. Now they are completely removed and a logEvent is added to show the action took place.

what needs to happen
- in the frontend it needs to have the same behaviour as the backend, so it needs to delete the comment completely from the timeline.

Possibly:
- Create a timelineEvent component for the comment deletion logEvent (in my opinion, that data does not help the user a lot)
- filter out comment deletion logEvents in frontend or backend

NRodriguezcuellar:
test comment

PR #254, opened 2022-06-22
requests timeline: synchronize comment deletion behaviour with backend 

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-requests/issues/251


alejandromumo:
Tests are failing, can you check if it's related with this PR @NRodriguezcuellar ?

Pineirin:
> Tests are failing, can you check if it's related with this PR @NRodriguezcuellar ?

@NRodriguezcuellar told me that tests are failing because he would need to fix them with each change, making it better for him to first get feedback on the implementation and then fix the tests.

Personally. I'm a bit confused by the PR and would thank a bit more context on the description telling what is supposed to do. For example: i don't now why expanded is needed here.

PR #256, opened 2022-06-27
comment deletion: synchronize behaviour with backend

NRodriguezcuellar:
When the user deletes a comment it will now be replaced by a deletion logEvent, in place of the deleted comment.
Screenshot from 2022-06-21 10-15-29

closes https://github.com/inveniosoftware/invenio-requests/issues/251

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
@@ -85,7 +85,7 @@ export const setPage = (page) => {
       payload: page,
     });
 
-    dispatch(fetchTimeline());
+    await dispatch(fetchTimeline());
   };
 };
 
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEventControlled/state/actions.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEventControlled/state/actions.js
@@ -12,6 +12,7 @@ import {
 } from "../../timeline/state/actions";
 import { payloadSerializer } from "../../api/serializers";
 import _cloneDeep from "lodash/cloneDeep";
+import { i18next } from "../../../../translations/invenio_requests/i18next";
 
 export const updateComment = ({ content, format, event }) => {
   return async (dispatch, getState, config) => {
@@ -46,7 +47,7 @@ export const deleteComment = ({ event }) => {
 
     dispatch({
       type: SUCCESS,
-      payload: _newStateWithDelete(event.id, getState().timeline.data),
+      payload: _newStateWithDelete(event.id, getState),
     });
 
     dispatch(setTimelineInterval());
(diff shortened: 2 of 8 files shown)


invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
```diff
@@ -107,8 +138,14 @@ const timelineReload = (dispatch, getState, config) => {
...
-    dispatch(fetchTimeline(true));
-    dispatch(setTimelineInterval());
+    const { firstFetch } = getState().timeline;
+
+    if (firstFetch) {
+      await dispatch(setLastPageOnLoad());
+    }
+
+    await dispatch(fetchTimeline(true));
+    await dispatch(setTimelineInterval());
```
alejandromumo, line 148: Are `await` needed here? E.g. is there any dependency between these `dispatch` calls? Applies to the whole function.

invenio_requests/assets/semantic-ui/js/invenio_requests/timelineCommentEventControlled/state/actions.js
```diff
@@ -72,11 +72,11 @@ const _newStateWithUpdate = (updatedComment, currentState) => {
 const _newStateWithDelete = (eventId, currentState) => {
   const timelineState = _cloneDeep(currentState);
 
-  const currentHits = timelineState.hits.hits;
-
-  const currentComment = currentHits.find((comment) => comment.id === eventId);
+  timelineState.hits.hits = timelineState.hits.hits.filter(
```
alejandromumo, line 75: It would be good to have some context here on why this is done like this. I understand it's due to ignoring `delete` events, but I was present in the demo.

Additionally, personally I am not a big fan of overwriting an array while iterating it. E.g. 

```
myVar.array = myVar.array.filter(...)
``` 

It is just a very personal remark and, since  it does not impact functionality, you can ignore this.

NRodriguezcuellar: I am always happy to hear what you think :smile: 

I should perhaps add a little description, what is it is meant to do is just find the item that was deleted and remove it from the current list of hits, in the current implementation i use filter() which should return a new array without the deleted item.

Pineirin: Maybe you could use the loadash remove function to make it cleaner: https://lodash.com/docs/4.17.15#remove (didn't test if it works in this particular case).

NRodriguezcuellar: i agree that it would be cleaner, although i also believe we should minimize our dependency on external libraries, for now i have changed it in another way, let me know what you guys think

tests/services/events/test_request_events_service.py
```diff
@@ -70,17 +70,13 @@ def test_simple_flow(
     deleted_item = request_events_service.delete(identity_simple, id_)
     assert deleted_item is True
     RequestEvent.index.refresh()
-    # assert that the comment was deleted and cannot be read anymore
-    with pytest.raises(NoResultFound):
-        request_events_service.read(identity_simple, id_)
```
Pineirin, line 75: If I understood properly, now you are updating the message to state the the comment was deleted. Of course then, it would never be completely deleted but, shouldn't you check that the message is that the appropriate one instead of just deleting the check?

NRodriguezcuellar: You're completely right, if you look a bit beneath this part you will see that we will check if there is a deleted comment

ntarocco (approved):
LGTM!

PR #262, opened 2022-07-04
User requests: remove requests with created status

Pineirin:
closes: inveniosoftware/invenio-requests#252

### DIFF
--- a/tests/services/requests/test_requests_service.py
+++ b/tests/services/requests/test_requests_service.py
@@ -121,18 +121,34 @@ def test_update_request(app, identity_simple, submit_request, requests_service):
 
 
 def test_search_user_requests(
-    app, identity_simple, identity_simple_2, users, submit_request, requests_service
+    app,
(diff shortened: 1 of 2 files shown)


ntarocco (commented):
Can we add a test?

tests/services/requests/test_requests_service.py
```diff
@@ -121,19 +121,34 @@ def test_update_request(app, identity_simple, submit_request, requests_service):
...
+    hits = requests_service.search_user_requests(identity_simple).to_dict()["hits"][
         "hits"
-    ]["hits"]
+    ]
 
     assert str(request_id) in [h["id"] for h in hits]
 
+    # requests with created status are not returned to the user
+    previous_hits = len(hits)
+    create_request(identity_simple)
```
kpsherva, line 145: do you refresh the index after creating the request ? because it might simply not be indexed yet in the test

PR #282, opened 2022-09-30
OpenSearch 2.x

max-moser:
=


invenio_requests/records/mappings/os-v2/requestevents/requestevent-v1.0.0.json
```diff
@@ -0,0 +1,43 @@
+{
+  "mappings": {
+    "dynamic": "strict",
```
ppanero, line 3: missing dynamic templates:

```
"dynamic_templates": [
        {
          "creators": {
            "path_match": "created_by.*",
            "mapping": {
              "type": "keyword"
            }
          }
        },
       ]
       ```
       
       Should not make tests fail anyway.

Issue #264, opened 2022-07-04
Implement feedback on mobile user requests.


Pineirin:
According to the feedback received, the following changes should be implemented on the mobile view of user requests:

- Center all content, of each of the requests.
- Move all the action buttons to a drop-down, the buttons should be hidden until the drop-down is opened. The proposed default drop-down is a vertical ellipsis.

kpsherva:
The components implemented are too complex in terms of overriding, there is a simpler way by using Media component from
https://www.npmjs.com/package/@artsy/fresnel?activeTab=readme, which should be implemented only in the RequestActions (invenio-requests module) without a need to provide overriden component. 
In addition we decided with @jennur to replace the buttons in the dropdown by simpler menu options (standard dropdown), therefore it should be aligned

PR #291, opened 2022-10-27
RequestsActions: fix merging issues

Pineirin:
* closes: inveniosoftware/invenio-requests#264

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/request/actions/RequestActions.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/request/actions/RequestActions.js
@@ -16,12 +16,12 @@ export const RequestActions = ({ request }) => {
   const { MediaContextProvider, Media } = AppMedia;
   return (
     <Overridable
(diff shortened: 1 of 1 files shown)


invenio_requests/assets/semantic-ui/js/invenio_requests/request/actions/RequestActions.js
```diff
@@ -21,7 +21,7 @@ export const RequestActions = ({ request }) => {
       actions={actions}
     >
       <MediaContextProvider>
-        <Media greaterThanOrEqual="tablet" className="fresnel-inline-block">
+        <Media greaterThanOrEqual="tablet" className="media-inline-block">
```
kpsherva, line 24: is it enough to use `.inline-block` ? I think that class already exists

kpsherva (commented):
please change the commit message, it is not informative on the change you are applying 
fe. requests: provide styling for action container

kpsherva (approved):

PR #292, opened 2022-10-27
Mm/bulk indexing

max-moser:
Register the requests/events services and indexers to the central registry provided by `invenio-records-resources` and `invenio-indexer`.
Also set more specific `indexer_queue_name` for the service configs.

To do:
* [ ] Utilize bulk indexing for `service.rebuild_index` rather than indexing every item individually


invenio_requests/views/__init__.py
```diff
@@ -8,10 +8,32 @@
...
 
+from flask import Blueprint
+
 from .api import create_request_events_bp, create_requests_bp
 from .ui import create_ui_blueprint
 
+blueprint = Blueprint("invenio-requests-ext", __name__)
+
+
+@blueprint.record_once
```
max-moser, line 19: this is basically the same approach as in [Invenio-RDM-Records](https://github.com/inveniosoftware/invenio-rdm-records/blob/e460ffa92aea052d7e9db1799e30e6971d8f10b3/invenio_rdm_records/views.py#L17-L32), and is done in order to delay the service/indexer registration until a point when all invenio modules have been initialized.
If this were done in `init_app(...)` in `ext.py`, it could happen that not all required modules (`invenio-indexer` and `invenio-records-resources`) have been initialized during the method's execution.

ppanero (approved):

zzacharo (approved):

PR #309, opened 2023-03-08
sercices: allow to set links_tpl

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2125

### DIFF
--- a/invenio_requests/services/requests/results.py
+++ b/invenio_requests/services/requests/results.py
@@ -57,6 +57,11 @@ class RequestItem(RecordItem):
         """Get links for this result item."""
         return self._links_tpl.expand(self._identity, self._request)
 
+    @links.setter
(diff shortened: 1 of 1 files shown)


invenio_requests/services/requests/results.py
```diff
@@ -57,6 +57,11 @@ def links(self):
         """Get links for this result item."""
         return self._links_tpl.expand(self._identity, self._request)
 
+    @links.setter
+    def links_tpl(self, links):
```
ppanero, line 61: minor: change the parameter name to `links_tpl` to make sure it is a template and not a dict, and avoid issues with subsequent calls to `links` (would fail to expand a dict)
```suggestion
    def links_tpl(self, links_tpl):
```

ppanero (approved):

PR #311, opened 2023-03-27
assets: add generic search components

kpsherva:
* refactor overridable components
* add contrib components: status labels, type labels
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2075


invenio_requests/assets/semantic-ui/js/invenio_requests/search/RequestStatusFilterComponent.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  componentDidMount() {
+    const { currentQueryState } = this.props;
+    const userSelectionFilters = currentQueryState.filters;
+    const openFilter = userSelectionFilters.find((obj) => obj.includes("is_open"));
+    if (openFilter) {
+      // eslint-disable-next-line react/no-did-mount-set-state
+      this.setState({
+        open: openFilter.includes("true"),
```
jrcastro2, line 31: Minor: Could this logic be moved to the constructor and set the state since the beginning, to avoid re-rendering?

invenio_requests/assets/semantic-ui/js/invenio_requests/contrib/labels/StatusLabel.js
```diff
@@ -9,59 +9,53 @@ import { Label, Icon } from "semantic-ui-react";
 import { i18next } from "@translations/invenio_requests/i18next";
 
 export const LabelStatusSubmit = (props) => {
```
jrcastro2, line 11: Minor: props can be removed. Also applies for the rest.
```suggestion
export const LabelStatusSubmit = (props) => {
```

invenio_requests/assets/semantic-ui/js/invenio_requests/search/RequestStatusFilterComponent.js
kpsherva: I would prefer not to touch any logic, since I just reorganised the component. I suggest you could create a ticket for this, if needed

PR #312, opened 2023-04-06
resolvers: use record-based resolvers and proxies

slint:
:heart: Thank you for your contribution!

### Description

Please describe briefly your pull request.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_requests/resolvers/requests.py
+++ b/invenio_requests/resolvers/requests.py
@@ -13,8 +13,9 @@ from invenio_records_resources.references.entity_resolvers import (
     RecordPKProxy,
     RecordResolver,
 )
+
 from ..records.api import Request, RequestEvent
-from ..services import RequestsServiceConfig, RequestEventsServiceConfig
+from ..services import RequestEventsServiceConfig, RequestsServiceConfig
 
 
 class RequestResolver(RecordResolver):
@@ -26,7 +27,6 @@ class RequestResolver(RecordResolver):
         """Initialize the default record resolver."""
         super().__init__(
             record_cls=Request,
-            # TODO: The request service is not reegistered in the service registry
             service_id=RequestsServiceConfig.service_id,
             type_key=self.type_id,
             proxy_cls=RecordPKProxy,
@@ -34,8 +34,7 @@ class RequestResolver(RecordResolver):
 
     def _reference_entity(self, entity):
         """Create a reference dict for the given request."""
-        id_ = entity.number if entity.number is not None else entity.id
-        return {"request": str(id_)}
+        return {self.type_key: str(entity.id)}
 
 
 class RequestEventResolver(RecordResolver):
@@ -47,7 +46,6 @@ class RequestEventResolver(RecordResolver):
         """Initialize the default record resolver."""
         super().__init__(
             record_cls=RequestEvent,
-            # TODO: The request service is not reegistered in the service registry
             service_id=RequestEventsServiceConfig.service_id,
             type_key=self.type_id,
             proxy_cls=RecordPKProxy,
--- a/invenio_requests/resolvers/requests.py
+++ b/invenio_requests/resolvers/requests.py
@@ -10,49 +10,49 @@
 """Resolver and proxy for requests."""
 
 from invenio_records_resources.references.entity_resolvers import (
-    EntityProxy,
-    EntityResolver,
+    RecordPKProxy,
+    RecordResolver,
 )
-from sqlalchemy.exc import StatementError
-from sqlalchemy.orm.exc import NoResultFound
-
-from ..records.api import Request
+from ..records.api import Request, RequestEvent
+from ..services import RequestsServiceConfig, RequestEventsServiceConfig
 
 
-class RequestProxy(EntityProxy):
-    """Resolver proxy for a Request entity."""
-
-    def _resolve(self):
-        """Resolve the Request from the proxy's reference dict."""
-        request_id = self._parse_ref_dict_id()
-        try:
-            return Request.get_record(request_id)
-        except StatementError as exc:
-            raise NoResultFound() from exc
-
-    def get_need(self):
-        """Return None since Needs are not applicable to requests."""
-        return None
-
-
-class RequestResolver(EntityResolver):
+class RequestResolver(RecordResolver):
     """Resolver for requests."""
 
     type_id = "request"
 
-    def matches_entity(self, entity):
-        """Check if the entity is a request."""
-        return isinstance(entity, Request)
+    def __init__(self):
+        """Initialize the default record resolver."""
+        super().__init__(
+            record_cls=Request,
+            # TODO: The request service is not reegistered in the service registry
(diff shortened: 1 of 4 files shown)


ntarocco (approved):

invenio_requests/resolvers/requests.py
```diff
@@ -10,49 +10,78 @@
...
+        )
 
     def _reference_entity(self, entity):
         """Create a reference dict for the given request."""
-        id_ = entity.number if entity.number is not None else entity.id
-        return {"request": str(id_)}
+        return {self.type_key: str(entity.id)}
+
+
+class RequestResultItemResolver(ResultItemResolver):
```
slint, line 62: minor: regarding the final usage of the `ResultItemResolver`, could we get away with not "pre-createing" these classes? Theoretically one can just initialize these as `ResultItemResolver(service_id="request")` and get away with it.

slint (commented):
LGTM, minor comment on if it's necessary to pre-create the `ResultItemResolver`-based classes.

invenio_requests/resolvers/requests.py
```diff
@@ -10,49 +10,49 @@
...
-        request_id = self._parse_ref_dict_id()
-        try:
-            return Request.get_record(request_id)
-        except StatementError as exc:
-            raise NoResultFound() from exc
+    def __init__(self):
+        """Initialize the default record resolver."""
+        super().__init__(
+            record_cls=Request,
+            # TODO: The request service is not reegistered in the service registry
```
slint, line 30: minor: We have to address this one first before merging. The solution is to do the classic [`Blueprint.record_once`](https://github.com/inveniosoftware/invenio-communities/blob/b9c21a7905ddef3a23ed0c24c75f711cca6af65d/invenio_communities/views/api.py#L16-L38) dance with the registry, for the requests and request events services

rekt-hard: just found it here:
https://github.com/inveniosoftware/invenio-requests/blob/45ca623c16d4ab2b81a4a7df26141123ac1edd60/invenio_requests/views/__init__.py#L19-L32

it's also setup here: https://github.com/inveniosoftware/invenio-requests/blob/45ca623c16d4ab2b81a4a7df26141123ac1edd60/setup.cfg#L52-L58

Could move it to `views/api.py` to be consistent with `invenio-communities`. Otherwise would just remove the TODO.

invenio_requests/resolvers/requests.py
```diff
@@ -10,49 +10,47 @@
...
-    def get_need(self):
-        """Return None since Needs are not applicable to requests."""
-        return None
+    def _reference_entity(self, entity):
+        """Create a reference dict for the given request."""
+        return {self.type_key: str(entity.id)}
 
 
-class RequestResolver(EntityResolver):
+class RequestEventResolver(RecordResolver):
```
kpsherva, line 40: if I remember correctly, this was not used anywhere until now, and renaming has no consequences we should mitigate elsewhere in the code?

invenio_requests/resolvers/requests.py
```diff
@@ -10,49 +10,47 @@
...
+            type_key=self.type_id,
+            proxy_cls=RecordPKProxy,
+        )
 
-    def get_need(self):
-        """Return None since Needs are not applicable to requests."""
-        return None
+    def _reference_entity(self, entity):
+        """Create a reference dict for the given request."""
+        return {self.type_key: str(entity.id)}
```
slint, line 37:
```suggestion
        id_ = entity.number if entity.number is not None else entity.id
        return {"request": str(id_)}
```

minor: this is the old behaviour for referencing a `Request`, though for some reason, I've never seen a reference dict it like this in practice in our instances... I would keep it maybe just for backwards compatibility.

invenio_requests/resolvers/requests.py
slint: (the diff renders a bit funny in this PR) It's not renamed, the original `RequestResolver` is still there. This just adds a resolver for `RequestEvent`s.

The change is that we change the base class to `RecordResolver` which minimized how many methods have to be overridden.

invenio_requests/resolvers/requests.py
slint: Yup, you're right I missed it completely... Ok, that explains actually why this worked without having a `service_id` in the service config. In that case I think you can remove the `service_id="requests"` and `service_id="request-evens"` parameters from the registration calls.

invenio_requests/resolvers/requests.py
rekt-hard: @kpsherva @ntarocco @zzacharo had a quick meeting about this. We did not find any reference in a quick search.
Additionally, the issue is with the `entity.number` being just a counter, starting from 1. When trying to resolve something like this (i.e. `"request": 4`), the record class will not be able to do so, since it is expecting a UUID value.

slint: Ok, that makes a lot of sense now. I thought there was some sort of `Request.number.resolve(...)` in place. In that case, we leave it as is :)

PR #313, opened 2023-04-11
translations: fix string concatenation and pluralizaton

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2162


ppanero (commented):
Can we/Does it make sense to use the `Trans` component?

PR #318, opened 2023-06-04
Make context vars in link generation customizable

max-moser:
This can be used to fix the wrong "self_html" links.

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/request/RequestMetadata.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/request/RequestMetadata.js
@@ -9,7 +9,7 @@ import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { Image } from "react-invenio-forms";
 import Overridable from "react-overridable";
-import { Divider, Header, Message } from "semantic-ui-react";
+import { Divider, Header, Icon, Message } from "semantic-ui-react";
 import { toRelativeTime } from "react-invenio-forms";
 import RequestStatus from "./RequestStatus";
 import RequestTypeLabel from "./RequestTypeLabel";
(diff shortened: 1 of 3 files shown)


invenio_requests/customizations/request_types.py
```diff
@@ -190,6 +190,16 @@ def marshmallow_schema(cls):
             current_requests._schema_cache[type_id] = cls._create_marshmallow_schema()
         return current_requests._schema_cache[type_id]
 
+    @classmethod
+    def link_config(cls, **context_values):
```
zzacharo, line 194: from the code below shouldn't this be `update_link_config`?

max-moser:
Expanded entities that are *not* communities or users will get rendered this way now:

image

In this case, it's a guest user with an external e-mail address for the guest access request.

invenio_requests/customizations/request_types.py
```diff
@@ -190,6 +190,19 @@ def marshmallow_schema(cls):
...
+        """Method for updating the context values when generating links.
+
+        WARNING: this method potentially mixes layers and might be a footgun;
+        it will likely be removed in a future release!
+
+        This method takes the already determined context values for the link as
+        keyword arguments and should return values that will be used to update the
+        original context values.
+        """
+        # FIXME/TODO this should be reworked into a service feature
```
max-moser, line 203: A rework in the service (config) might take advantage of the fact that the var update for links takes the record/object as argument...
Relevant pointers:
* https://github.com/inveniosoftware/invenio-rdm-records/pull/1284/files#diff-640bfb3d54c64abe82cf753d8ffe63ddc5735cc6cca4388f6c9fc8c339f04102R357
* https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/base/links.py#L111
* https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/services/requests/config.py#L79

Maybe it will require making the `service.config.links_item` configurable (similar to the permission policy), or maybe it can be added to the `RequestLink` class somehow.
Needs some more investigation :detective:

zzacharo (approved):

invenio_requests/assets/semantic-ui/js/invenio_requests/request/RequestMetadata.js
```diff
@@ -33,15 +33,24 @@ const Community = ({ community }) => (
...
+    <Icon name="mail" className="mr-5" />
+    <span>{i18next.t("External email")}: {email.id}</span>
+  </div>
+);
 
-const UserOrCommunity = ({ userData, details }) => {
+const EntityDetails = ({ userData, details }) => {
   const isUser = "user" in userData;
   const isCommunity = "community" in userData;
+  const isEmail = "email" in userData;
```
ntarocco, line 46: Is it always external when `email`? If yes, very minor:
```suggestion
  const isExternalEmail = "email" in userData;
```
and same for methods/components.

ntarocco (approved):

invenio_requests/assets/semantic-ui/js/invenio_requests/request/RequestMetadata.js
max-moser: Yes, indeed – since there's no notion of "internal" emails (unless they are associated with a user account), they are necessarily "external".
Fixed!

zzacharo (approved):

PR #335, opened 2023-07-19
moderation: added request type and service

alejandromumo:
closes https://github.com/inveniosoftware/invenio-requests/issues/334

NEEDS:

- https://github.com/inveniosoftware/invenio-users-resources/pull/91


invenio_requests/services/user_moderation/user_moderation.py
```diff
@@ -0,0 +1,56 @@
...
+class ApproveUserAction(actions.AcceptAction):
+    """Represents an accept action used to aprove an user."""
+
+    def execute(self, identity, uow):
+        """Executes aprove action."""
+        # TODO add specific user block actions
+        super().execute(identity, uow)
+
+
+class UserModeration(RequestType):
```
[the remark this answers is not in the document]
alejandromumo: That is a good suggestion, I was not sure where to put the request type. 
thanks!

invenio_requests/services/user_moderation/user_moderation.py
```diff
@@ -0,0 +1,56 @@
...
+    """Request to moderate an user."""
+
+    type_id = "user-moderation"
+    name = _("User moderation")
+
+    creator_can_be_none = False
+    topic_can_be_none = False
+    allowed_creator_ref_types = ["user"]
+    allowed_receiver_ref_types = ["user"]
```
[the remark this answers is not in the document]
alejandromumo: Yes, I need to review permissions in the next iteration.  I will implement the actions (accept, decline) and configure the permissions. 

About the invenio-administration generator, I tried to use it in user's service but it was not working. Namely, if the generators of an action (e.g. `search`) have more than the administration (`Administration()`) then it would disappear and raise permission errors. As far as I debugged, this was happening [here](https://github.com/inveniosoftware/invenio-records-permissions/blob/master/invenio_records_permissions/policies/base.py#L88). 

In the end, I am using `SystemProcess()`. But I do agree we need to review this and understand if that's good enough for the user moderation actions.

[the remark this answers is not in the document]
ntarocco: When you get there, maybe we can discuss it quickly, having a look to some code. In any case, `Administration` role must have access to the request, otherwise these users cannot perform searches and see it in the admin panel.

alejandromumo:
Tests are passing locally. The are failing on CI because it needs another PR (users-resources).

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,70 @@
...
+
+from invenio_access.permissions import system_identity, system_user_id
+from invenio_i18n import gettext as _
+
+from invenio_requests.resolvers.registry import ResolverRegistry
+from invenio_requests.services.user_moderation.errors import InvalidCreator
+from invenio_requests.services.user_moderation.user_moderation import UserModeration
+
+
+class UserModerationRequestService:
```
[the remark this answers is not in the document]
alejandromumo: I had to go a bit further and implement this service as a `Service`, mainly due to permissions and search.

invenio_requests/services/user_moderation/user_moderation.py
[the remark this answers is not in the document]
alejandromumo: I created an entity resolver for this type so the user moderator (And system process) can execute actions.

tests/services/user_moderation/test_user_moderation_service.py
```diff
@@ -0,0 +1,71 @@
...
+import pytest
+from flask_principal import Need
+from invenio_access.permissions import system_identity, system_user_id
+from invenio_access.utils import get_identity
+
+from invenio_requests.proxies import current_user_moderation_service
+from invenio_requests.services.user_moderation.errors import InvalidCreator
+
+
+def test_request_moderation(app, users, identity_simple):
```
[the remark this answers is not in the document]
alejandromumo: I added new tests, both for the resource and service.

invenio_requests/customizations/user_moderation/user_moderation.py
```diff
@@ -0,0 +1,69 @@
...
+
+class UserModeration(RequestType):
+    """Request to moderate an user."""
+
+    type_id = "user-moderation"
+    name = _("User moderation")
+
+    creator_can_be_none = False
+    topic_can_be_none = False
+    allowed_creator_ref_types = ["user_moderation"]
```
alejandromumo, line 50: can be system process or a user moderator

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,112 @@
...
+
+from invenio_requests.customizations.user_moderation.user_moderation import (
+    UserModeration,
+)
+from invenio_requests.proxies import current_request_type_registry
+from invenio_requests.resolvers.registry import ResolverRegistry
+from invenio_requests.services.user_moderation.errors import InvalidCreator
+
+
+class UserModerationRequestService(Service):
```
alejandromumo, line 22: Implemented as a `Service` due to permissions and search filters. 

We need to discuss what happens if this service is bypassed (e.g. executing actions directly in the requests service).

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,112 @@
...
+
+        return self.requests_service.read(
+            identity=system_identity, id_=request_id, **kwargs
+        )
+
+    def moderate(self, identity, request_id, action, data=None):
+        """Moderates a user."""
+        self.require_permission(identity, "moderate")
+
+        return self.requests_service.execute_action(
```
alejandromumo, line 47: permissions for each action will be checked again using the generic requests service.

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def request_type_cls(self):
+        """User moderation request type."""
+        return current_request_type_registry.lookup(UserModeration.type_id)
+
+    @unit_of_work()
+    def request_moderation(self, identity, topic, data=None, uow=None, **kwargs):
+        """Creates a UserModeration request and submits it."""
+        role = Role.query.filter(Role.name == MODERATION_ACTION_NAME).one_or_none()
```
lnielsen, line 34: Minor: The fact is it can change afterwards, i.e. the role which receives the request is not guaranteed to have permissions when they press accept/decline. Also it assumes that only a role can hold the action, but a user can also hold the action.

Logic is almost the same, but naming and concept is different:

```python
role = Role.query.filter(Role.name == REQUEST_MODERATOR_ROLEID ).one_or_none()
```

Get a "hardcoded role" from the config so each instance can set it to whatever they want.

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,59 @@
...
+        assert role, _("Moderation role must exist to enable user moderation requests.")
+
+        data = data or {}
+
+        # For user moderation, topic is the user to be moderated
+        topic = {"user": str(topic)}
```
lnielsen, line 40: Question: Should the service method hardcode that we want a user and check that the user exists? Might be this is the end-result but not very clear from here, so if the case, then add a comment.

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Receiver can be configured, by default send the request to users with moderation role
+        receiver = {"group": role.name}  # TODO to be changed to role id
+        creator = {"group": role.name}  # TODO to be changed to role id
```
lnielsen, line 44: Question: Should this be the system user instead?

invenio_requests/customizations/user_moderation/user_moderation.py
```diff
@@ -0,0 +1,55 @@
...
+    creator_can_be_none = False
+    topic_can_be_none = False
+    allowed_creator_ref_types = ["group"]
+    allowed_receiver_ref_types = ["group"]
+    allowed_topic_ref_types = ["user"]
+
+    available_actions = {
+        "delete": actions.DeleteAction,
+        "submit": actions.SubmitAction,
+        "create": actions.CreateAction,
```
zzacharo, line 50: nit: can we do something like 
```suggestion
        **RequestType.available_actions,
        ...
```
?

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,62 @@
...
+#
+# Invenio-Requests is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""User moderation service."""
+from flask import current_app
+from invenio_accounts.models import Role
+from invenio_i18n import gettext as _
+from invenio_records_resources.services.uow import unit_of_work
+
+from invenio_requests.customizations.user_moderation.user_moderation import (
```
zzacharo, line 13:
```suggestion
from invenio_requests.customizations.user_moderation import (
```

invenio_requests/services/user_moderation/service.py
```diff
@@ -0,0 +1,62 @@
...
+    @unit_of_work()
+    def request_moderation(self, identity, user_id, data=None, uow=None, **kwargs):
+        """Creates a UserModeration request and submits it."""
+        REQUESTS_MODERATION_ROLE_ID = current_app.config.get(
+            "REQUESTS_MODERATION_ROLE_ID"
+        )
+        role = Role.query.filter(Role.name == REQUESTS_MODERATION_ROLE_ID).one_or_none()
+        assert role, _("Moderation role must exist to enable user moderation requests.")
+
+        data = data or {}  # TODO is this being validated?
```
zzacharo, line 40: TODO relevant?

zzacharo (approved):

PR #343, opened 2023-08-10
access request: update guest request payload

anikachurilova:
* add an overridable for the timeline header
* inject permissions prop
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1395

NOTE: the permissions for request.payload updation should be handled in separate PR (this TODO https://github.com/inveniosoftware/invenio-requests/pull/343/files#diff-4b568e443f92c06d4ea44587cf21b1134cfaff1f5bf4ceb88076ae894c452f0fR44)
The task is already created: https://github.com/inveniosoftware/invenio-rdm-records/issues/1402

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/requestsAppInit.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/requestsAppInit.js
@@ -12,12 +12,20 @@ import {
 import { i18next } from "@translations/invenio_requests/i18next";
 import React from "react";
 import ReactDOM from "react-dom";
+import { overrideStore } from "react-overridable";
+import { InvenioRequestsApp } from "./InvenioRequestsApp";
 import {
   RequestAcceptButton,
   RequestCancelButton,
   RequestDeclineButton,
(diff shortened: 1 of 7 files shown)


invenio_requests/assets/semantic-ui/js/invenio_requests/requestsAppInit.js
```diff
@@ -72,12 +74,17 @@ const overriddenComponents = {
...
+for (const [key, value] of Object.entries(overriddenComponents)) {
+  overrideStore.add(key, value);
+}
+
 ReactDOM.render(
   <InvenioRequestsApp
     request={request}
     defaultQueryParams={defaultQueryParams}
-    overriddenCmps={overriddenComponents}
+    overriddenCmps={overrideStore.getAll()}
```
ntarocco, line 85: I would not do that to avoid, I am not sure in which case, replacing an already defined id in the global store.
We should prefer the pattern:
```
{ ...defaultComponents, ...overriddenComponents },
```
instead.

ntarocco (approved):

PR #348, opened 2023-08-18
wip: added administration panel ui skeleton

alejandromumo:
closes https://github.com/inveniosoftware/invenio-requests/issues/347

# Important (from Karolina):

verify https://github.com/inveniosoftware/invenio-requests/pull/348/files#r1319860338

# Archive comments (Manuel):
This is the first iteration of the UI.

Implemented:

- View registration
- Approve / Block button and actions (HTTP request is executed when the buttons are clicked)
- Basic search / facets functionalities

Missing:
 - Missings facets
 - Missing user information on the list, like `user.email` (maybe we can use `?expand=True`)
 - Many other things :)

Below some ideas on how to address the missing parts (tested locally):

- User information can be retrieved by querying the requests resource using the parameter `expand=true`. E.g:

```python
    # invenio_requests.administration.user_moderation.user_moderation.py
    def get_api_endpoint(self):
        """Get search API endpoint.

        Filters only 'user-moderation' requests.
        """
        ...
        if not self.api_endpoint.startswith(api_url_prefix):
            return f"{api_url_prefix}{slash_tpl}{self.api_endpoint}?expand=true&q={filter_q}"
```

If the `User` entity should expand with more information, it should be implemented in its entity resolver ([here](https://github.com/inveniosoftware/invenio-users-resources/blob/master/invenio_users_resources/entity_resolvers.py#L61))

Moreover, the `topic`  field is not expandable right now. It must be enabled in `invenio-requests`:

```python
     # invenio_requests.services.requests.service.py
    @property
    def expandable_fields(self):
        """Get expandable fields."""
        return [
            EntityResolverExpandableField("created_by"),
            EntityResolverExpandableField("receiver"),
            EntityResolverExpandableField("topic"),
        ]
```

Then enable it in the administration view:

```python
    # invenio_requests.administration.user_moderation.user_moderation.py
    item_field_list = {
        "id": {
            "text": "ID",
            "order": 1,
        },
        "expanded.topic.email": {"text": "Email", "order": 5},  # like this
        "created": {"text": "Created", "order": 2},
        "updated": {"text": "Updated", "order": 3},
        "is_open": {"text": "Open", "order": 4},
    }
```

There is an issue with with `GroupProxy` (entity resolvers too), the following method is missing there (otherwise it fails while expanding):

```python
    def ghost_record(self, value):
        """Return default representation of not resolved group."""
        return {}
```


kpsherva:
Screenshot 2023-09-04 at 19 30 36

invenio_requests/services/requests/service.py
```diff
@@ -58,6 +58,8 @@ def expandable_fields(self):
         return [
             EntityResolverExpandableField("created_by"),
             EntityResolverExpandableField("receiver"),
+            # TODO to be verified
+            EntityResolverExpandableField("topic"),
```
kpsherva, line 62: I find this a bit controversial, since I it could open a path to resolving restricted records as topic of a request, however, without this we won't be able to see the user (which is under the topic in moderation requests)

slint: I agree, it's a bit touchy, since different request types, might hold heavier or even restricted information... Is there any request-type-specific service schema permissions-based filtering that happens afterwards?

PR #350, opened 2023-08-21
Field permissions

kpsherva:


invenio_requests/customizations/request_types.py
```diff
@@ -167,15 +177,12 @@ def _create_marshmallow_schema(cls):
...
 
-        # Raise on invalid payload keys
-        class PayloadBaseSchema(ma.Schema):
-            class Meta:
-                unknown = ma.RAISE
-
         # If a payload schema is defined, add it to the request schema
         if cls.payload_schema is not None:
+            # TODO probably could be improved, probably in a wrong place
+            cls._create_payload_cls()
```
zzacharo, line 183: check my comment https://github.com/inveniosoftware/invenio-rdm-records/pull/1410#discussion_r1303086158

zzacharo: discussed to drop this idea as it is not suited for dynamic class loading

PR #357, opened 2023-09-12
permissions: fix delete bug

jrcastro2:


invenio_requests/services/requests/service.py
```diff
@@ -202,7 +202,7 @@ def delete(self, identity, id_, uow=None):
         # self.check_revision_id(request, revision_id)
 
         # check permissions
-        self.require_permission(identity, f"delete", request=request)
+        self.require_permission(identity, f"action_delete", request=request)
```
jrcastro2, line 205: Comment: Feels a bit redundant and confusing to have a delete method and an `execute_action` that allows the `delete` action. The difference I can see is that the `delete` method does run the components for that action (and none of these components do anything on delete). I didn't change anything yet as I would like to have other opiniones but I would move the `run_components` to the `execute_action` and remove the `delete` func

zzacharo (approved):

PR #358, opened 2023-09-12
ui: support community manage record request facets and labels

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/243

Screenshot 2023-09-12 at 16 15 59


invenio_requests/assets/semantic-ui/js/invenio_requests/contrib/index.js
slint, line 1: comment/shelve: I'm somewhat confused that we include in `invenio-requests` mappings/listing of request types that are originally defined in other modules. I could imagine it being difficult to track these down when e.g. trying to implement a new request type.

I guess it's because there's this mix of Python/JS-land which makes it difficult to dynamically define things from config. In any case, something to shelve and revisit.

PR #375, opened 2024-01-30
http headers: use and adjust vnd.inveniordm.v1+json http accept header

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/598

### DIFF
--- a/invenio_requests/customizations/event_types.py
+++ b/invenio_requests/customizations/event_types.py
@@ -104,9 +104,9 @@ class EventType:
         """Create a schema for the entire request including payload."""
         type_id = cls.type_id
         if type_id not in current_requests._events_schema_cache:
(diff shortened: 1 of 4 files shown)


invenio_requests/customizations/event_types.py
```diff
@@ -104,9 +104,9 @@ def marshmallow_schema(cls):
         """Create a schema for the entire request including payload."""
         type_id = cls.type_id
         if type_id not in current_requests._events_schema_cache:
-            current_requests._events_schema_cache[
-                type_id
-            ] = cls._create_marshmallow_schema()
+            current_requests._events_schema_cache[type_id] = (
```
anikachurilova, line 107: This was failing because of the new version of `black`

PR #398, opened 2024-08-06
editor: prevent rerender on focus loss

max-moser:
Basically the same as https://github.com/inveniosoftware/react-invenio-forms/pull/249, but ported to text inputs in requests.
Closes https://github.com/inveniosoftware/invenio-requests/issues/399

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,7 +63,7 @@ jobs:
           pip install ".[$EXTRAS]"
           pip freeze
           docker --version
-          docker-compose --version
+          docker compose version
       - name: Run tests
         run: |
           ./run-tests.sh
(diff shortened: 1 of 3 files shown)


Samk13 (approved):

max-moser:
I just linked this PR to a new issue and tested the commit (cherry-picked on top of v4.1.1), and the issue with the "action + comment" editor is still happening.
I think it is indeed linked to the `onChange` handler, as that's pretty much the only spot where the string "Add comment (optional)" is defined, other than translations.

Samk13:
> I just linked this PR to a new issue and tested the commit (cherry-picked on top of v4.1.1 for my local v12), and the issue with the "action + comment" editor is still happening. I think it is indeed linked to the `onChange` handler, as that's pretty much the only spot where the string "Add comment (optional)" is defined, other than translations.

It shouldn't work if you only cherry-pick it. You need to watch the assets during testing to see the changes.
I am testing all editors here on V12, and it seems to work as expected, so I am not sure what is not working in your case.
editors-test

PR #413, opened 2024-11-14
requests: manage sending notifications

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/203

### DIFF
--- a/invenio_requests/services/requests/service.py
+++ b/invenio_requests/services/requests/service.py
@@ -224,7 +224,9 @@ class RequestsService(RecordService):
         action_obj.execute(identity, uow)
 
     @unit_of_work()
-    def execute_action(self, identity, id_, action, data=None, uow=None, expand=False):
+    def execute_action(
+        self, identity, id_, action, data=None, uow=None, expand=False, **kwargs
+    ):
         """Execute the given action for the request, if possible.
 
         For instance, it would be not possible to execute the specified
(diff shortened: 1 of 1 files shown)


invenio_requests/services/requests/service.py
```diff
@@ -244,7 +246,7 @@ def execute_action(self, identity, id_, action, data=None, uow=None, expand=Fals
             raise CannotExecuteActionError(action)
 
         # Execute action and register request for persistence.
-        action_obj.execute(identity, uow)
+        action_obj.execute(identity, uow, **kwargs)
```
alejandromumo, line 249: I think this is backwards compatible, but I am a bit worried so it's better to double-check.

Other actions have their signature unchanged (e.g. [this action](https://github.com/inveniosoftware/invenio-communities/blob/006f2f4e0077a900a02e35d839586005f84e6608/invenio_communities/subcommunities/services/request.py#L21)) . Can you just confirm the change is compatible?

alejandromumo (commented):
Reviewed with @carlinmack

invenio_requests/services/requests/service.py
anikachurilova: I tested on [this one](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/requests/community_submission.py#L101) and I confirm that everything works fine 🚀

Issue #428, opened 2025-03-13
Assign reviewers to a request


zzacharo:
## Context

As a community curator, I want to allow specific users/groups per request to be able to comment on a specific request. These users should not be able to change the status of the request or modify the record.

## Requirements

- We need to be able to assign reviewers to a specific request. 
- A reviewer can be either a user or a group. 
- Once added, a reviewer can
  - comment on a request
  - to explore: indicate if they approve a request. In some contexts, e.g CERN institutional repository, the approval action leads to confusion because of the existence of the "Accept and publish"

## Existing prototype

In the linked PRs, there is some prework done:

- New `reviewers` field added in the request jsonschema (missing to be added in the opensearch mappings)
- The entity reference/resolution classes were duplicated to support array of entities. They should be merged or refactored to utilize the single entity reference classes.
- (**This can be splitted in a separate issue**) Evaluate if we need to store the `opiniated` review in a separate field ala GH.
  - If not, remove the field `lastOpiniatedReview` (name to be checked)
  - If yes, change the code to support a list of opiniated reviews in accordance with the reviewers 

## Future Ideas/Discussions

- https://github.com/inveniosoftware/invenio-requests/pull/431#discussion_r2033458018
- Add a notification for review
- Add role name in from of the user name in each comment ex: `User 1 (reviewer)`

## Tasks

- [x] Config for disabling groups
- [x] feature flag for entire feature
- [x] UI/UX - Fix UI when no reviewer is selected
- [x] Fix UI with larger names
- [x] Add timeline event when reviewer added/removed
- [ ] Fix access for the reviewer after removed (TBD)
- [x] Limit people who can add a reviewer (TBD)
- [ ] Add tests

PR #440, opened 2025-04-11
Extra context for permission checks

xsenyaaax:
### Description

At CESNET we extend requests' services to have them as first class citizens (we have REST API to create requests in a generic way). In order to do so, we need to extend permission checks so that they not only include requests but also create/update data and additional parameters from context.

This pull request:

- add kwargs to service methods
- passes the kwargs data and additional context to permission checks

This change is 100% backwards compatible and in RDM **does not** cause any differences in permission handling 

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_requests/services/events/service.py
+++ b/invenio_requests/services/events/service.py
@@ -4,6 +4,7 @@
 # Copyright (C) 2021-2022 Northwestern University.
 # Copyright (C) 2021-2022 TU Wien.
 # Copyright (C) 2023 Graz University of Technology.
+# Copyright (C) 2026 CESNET z.s.p.o.
 #
 # Invenio-Requests is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -77,6 +78,7 @@ class RequestEventsService(RecordService):
         expand=False,
         notify=True,
         parent_id=None,
+        **kwargs,
     ):
         """Create a request event (top-level or reply).
 
@@ -87,13 +89,29 @@ class RequestEventsService(RecordService):
         :param parent_id: Optional parent event ID for replies.
         """
         request = self._get_request(request_id)
-        self.require_permission(identity, "read", request=request)
+        self.require_permission(
+            identity,
+            "read",
+            request=request,
+            data=data,
+            event_type=event_type,
+            notify=notify,
+            **kwargs,
+        )
         try:
             # If the event is a log, we don't check for permissions to not block logs creation
             if event_type.type_id != LogEventType.type_id:
                 # Check permission based on whether this is a reply or top-level comment
                 permission = "reply_comment" if parent_id else "create_comment"
-                self.require_permission(identity, permission, request=request)
+                self.require_permission(
+                    identity,
+                    permission,
+                    request=request,
+                    data=data,
+                    event_type=event_type,
+                    notify=notify,
+                    **kwargs,
+                )
         except PermissionDeniedError:
             if current_app.config.get(
                 "REQUESTS_LOCKING_ENABLED", False
@@ -182,12 +200,12 @@ class RequestEventsService(RecordService):
             request=request,
         )
 
-    def read(self, identity, id_, expand=False):
+    def read(self, identity, id_, expand=False, **kwargs):
         """Retrieve a record."""
         event = self._get_event(id_)
         request = self._get_request(event.request_id)
 
-        self.require_permission(identity, "read", request=request)
+        self.require_permission(identity, "read", request=request, **kwargs)
 
         return self.result_item(
             self,
@@ -203,13 +221,20 @@ class RequestEventsService(RecordService):
         )
 
     @unit_of_work()
-    def update(self, identity, id_, data, revision_id=None, uow=None, expand=False):
+    def update(
+        self, identity, id_, data, revision_id=None, uow=None, expand=False, **kwargs
+    ):
         """Update a comment (only comments can be updated)."""
         event = self._get_event(id_)
         request = self._get_request(event.request.id)
         try:
             self.require_permission(
-                identity, "update_comment", request=request, event=event
+                identity,
+                "update_comment",
+                request=request,
+                event=event,
+                data=data,
+                **kwargs,
             )
         except PermissionDeniedError:
             if current_app.config.get(
@@ -275,7 +300,7 @@ class RequestEventsService(RecordService):
         )
 
     @unit_of_work()
-    def delete(self, identity, id_, revision_id=None, uow=None):
+    def delete(self, identity, id_, revision_id=None, uow=None, **kwargs):
         """Delete a comment (only comments can be deleted)."""
         event = self._get_event(id_)
         request_id = event.request_id
@@ -283,7 +308,7 @@ class RequestEventsService(RecordService):
 
         # Permissions
         self.require_permission(
-            identity, "delete_comment", request=request, event=event
+            identity, "delete_comment", request=request, event=event, **kwargs
         )
         self.check_revision_id(event, revision_id)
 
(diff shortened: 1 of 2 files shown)


invenio_requests/services/events/service.py
```diff
@@ -55,7 +56,15 @@ def create(
         :param dict data: Input data according to the data schema.
         """
         request = self._get_request(request_id)
-        self.require_permission(identity, "create_comment", request=request)
+        self.require_permission(
+            identity,
+            "create_comment",
+            request=request,
+            data=data,
```
zzacharo, line 63: Do you run specific permission checks on comment actions based on request and event_type? I am trying to understand if we don't overpass kwargs down to `require_permission`. Also, I would be interested to know on how you use them in your use cases :)

mesemus: Hi Zach,

General intro (sorry, it will be a bit longer):

What we do in our overlay is that each community can have one or more associated workflows (one of them being "default"). The workflow handles the exact deposition/record lifecycle, regardless of whether the record is in draft or published state.

When a record is created within a community (we don't allow creating records outside of a community), we take the default workflow from the community and set a `workflow` attribute (string) on the record's parent. This `workflow` then identifies the exact deposition/record lifecycle workflow.

Additionally, each record has a `state` field (initially set to `draft`), and record-level permissions use it for checks.

Workflows are defined in Python (invenio.cfg) as a combination of record permissions and requests permissions:

```python
WORKFLOWS = {
    "default": Workflow(
        label=_("Default workflow for normal communities"),
        permission_policy_cls=DefaultWorkflowPermissions,
        request_policy_cls=DefaultWorkflowRequests,
    ),
    ...
}
```

Example from the record permission policy part:

```python
# piece of DefaultWorkflowPermissions

can_read = [
      IfInState(
          "published",
          then_=[
              IfRestricted(
                  "record",
                  then_=[PrimaryCommunityMembers()],
                  else_=[AnyUser()],
              )
          ],
      ),
      IfInState(
          "retracting",
          then_=[
              RecordOwner(),
              PrimaryCommunityRole("curator")
          ],
      ),
]
```

When permission is checked on a record (for example, `can_read`), the record's permission policy will look at the record's parent, find the workflow ID, resolve it to the record permission policy associated with the workflow, and execute `can_read` on it. There, `IfInState` looks at the state field in the record metadata, and if the record is in `retracting` state, only the record owner and curator are given access.

**Now, requests**

The second part of the workflow defines who can create requests, when they can create them, what events can occur on those requests, who can accept/decline them, etc. It also defines escalations if the request isn't handled within a specified time interval.

Example:

```python
class GenericCommunityWorkflowRequests(WorkflowRequestPolicy):
    delete_published_record = WorkflowRequest(
        requesters=[
            IfNotHarvested(
                then_=IfInState(
                    "published",
                    then_=[
                        RecordOwners(),
                        PrimaryCommunityRole("curator"),
                        PrimaryCommunityRole("owner"),
                    ],
                ),
                else_=[SystemProcess()],
            )
        ],
        recipients=[
            IfRequestedBy(
                requesters=[
                    PrimaryCommunityRole("curator"),
                    PrimaryCommunityRole("owner"),
                ],
                then_=[AutoApprove()],
                else_=[PrimaryCommunityRole("curator")],
            )
        ],
        transitions=WorkflowTransitions(
            submitted="retracting",
            declined="published",
            accepted="deleted",
            cancelled="published",
        ),
        events={
            "event_type": WorkflowEvent(
                submitters=[
                    # generators that can use the topic, request, request_type, event_type
                ],
            )
        }
        escalations=[
            WorkflowRequestEscalation(
                after=timedelta(days=21),
                recipients=[
                    PrimaryCommunityRole("owner"),
                ],
            )
        ],
    )
```

Imagine that the owner of `/api/records/abc` would like to delete the record. With workflows, the user directly creates a request for that:

```rest
POST /api/records/abc/requests/delete_published_record

{...potentially some payload here, such as deletion reason...}
```

This in turn calls Invenio Request Service's create method, passing the payload and *additionally the `abc` record in kwargs.*

We've replaced the `can_create` permission of the Invenio Requests Service with:

```python
    can_create = [
        SystemProcess(),
        RequestCreatorsFromWorkflow(),
    ]
```

For the POST above: `RequestCreatorsFromWorkflow` takes the `abc` record from the kwargs, looks up its workflow, finds `delete_published_record` in the workflow's requests configuration, and uses the `requesters` field.

If the user has permission, it will use the `recipients` property from the configuration above to determine who should be placed in the `recipient` field of the newly created request. An `AutoApprove` above is a special case - the request is accepted automatically.

When the request is submitted, the transitions part will automatically change the state field on the record to `retracting`, causing that only the owner and curator will have access to the record at this stage.

Finally, the `escalations` part will check in a Celery task that the request hasn't been handled for x days, will reassign the recipient of the request to a new person, and send notifications.

For events, it works the same way, using the `submitters` on the `WorkflowEvent`. We use different event types (not just comment events) during the lifecycle of the request - for example, if we have a repository with sensitive data where access to files is controlled by a data access committee, we'll use a special event type to record the committee's decision in a structured way.

*Final note:* we are not the ones that write the workflow policies - it is the owners of the repositories built on our overlay. That's why we would like to pass as complete a context to permissions (and components as well) as possible to enable them writing their possibly complex permission generators.

So that's why we need to have those kwargs and extra context :) We've been using this from our own fork for the last year or so. If it gets into Invenio main, we'll be able to remove our fork. If it doesn't fit RDM (for example, for security/API compactness considerations), I understand - no harm done, we'll just keep our patch.

zzacharo: Hey @mesemus thanks a lot for the amazing explanation! I see you have some really advanced customization which might be also interesting for the core. I understand now a bit better your use case but now I am intrigued to know a bit more (maybe a demo?) of your use case examples. Have you also changed the UI to accomodate all these different deposition workflows?

I am interested because I have started investigating (we are in the phase of conducting some interviews with our CERN stakeholders) potential solutions/improvements on the publication workflow we support currently on RDM. For example, we are working now to add "[Reviewers](https://github.com/inveniosoftware/invenio-requests/issues/428)" as a first class citizen of the request and in parallel thinking on how to improve the commenting capabilities on a request.

I will be off until 22nd of April but I would be interested to see in a demo your use case and consider if there is any common pathway to integrate fully or partially the customizable publication workflows.

mesemus: @zzacharo sure, let's move this to Discord to find a time slot.

zzacharo (approved):
@mesemus my only suggestion for this is to maybe rename kwargs added in the signature of the methods to `permission_context` or similar so that is more clear that the  extra kwards are permission oriented. Unless @slint has a different opinion, I think after the rename we can go ahead and merge this across the modules :)

utnapischtim (approved):

PR #466, opened 2025-07-25
requests: fix request UI when no receiver is allowed

yashlamba:
We allow the receiver to be none on the backend, but since it was not used until now, the frontend was probably to tested for it. We are using this for the `record-deletion` request type and came across the issue.

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/search/MobileRequestItem.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/search/MobileRequestItem.js
@@ -76,7 +76,7 @@ export const MobileRequestItem = ({
             {creatorName}
           </small>
           <small className="block rel-mt-1">
-            {result.receiver.community && result.expanded?.receiver.metadata.title && (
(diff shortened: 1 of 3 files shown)


invenio_requests/assets/semantic-ui/js/invenio_requests/search/MobileRequestItem.js
```diff
@@ -76,16 +76,19 @@ export const MobileRequestItem = ({
...
-            )}
+            {result.receiver?.community &&
+              result.expanded?.receiver?.metadata.title && (
+                <>
+                  <Icon
+                    className="default-margin"
+                    name={getUserIcon(result.expanded?.receiver)}
+                  />
+                  <span className="ml-5">
+                    {result.expanded?.receiver.metadata.title}
```
carlinmack, line 87: minorest of nits, but we know we have this here

```suggestion
                    {result.expanded.receiver.metadata.title}
```

carlinmack (commented):
lgtm!

PR #470, opened 2025-08-05
global: add `Request.last_reply` systemfield

slint:
* Adds a computed and search-indexed field for getting the last reply
  (i.e. the last comment event) for a request.

### DIFF
--- a/invenio_requests/services/schemas.py
+++ b/invenio_requests/services/schemas.py
@@ -14,67 +14,14 @@ from datetime import timezone
 
 from invenio_records_resources.services.records.schema import BaseRecordSchema
 from marshmallow import (
-    INCLUDE,
     RAISE,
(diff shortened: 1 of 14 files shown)


invenio_requests/services/schemas.py
```diff
@@ -14,21 +14,47 @@
...
+class EventTypeMarshmallowField(fields.Str):
+    """Serializes event type from `EventType` to string."""
+
+    def _serialize(self, value, attr, data, **kwargs):
+        if isinstance(value, EventType):
+            value = value.type_id
+        return value
+
+
+class RequestEventSchema(BaseRecordSchema):
```
slint, line 36: Similar to the API classes, this reordering is so that we can reference this class below in the `RequestSchema.last_reply` nested field

PR #474, opened 2025-08-21
reviewers: fix removing the last reviewer

zzacharo:
closes https://github.com/CERNDocumentServer/cds-rdm/issues/548

### DIFF
--- a/invenio_requests/services/requests/components.py
+++ b/invenio_requests/services/requests/components.py
@@ -58,30 +58,46 @@ class RequestReviewersComponent(ServiceComponent):
     """Component for handling request reviewers."""
 
     def _reviewers_updated(self, previous_reviewers, new_reviewers):
-        """Determine the event type based on reviewers added or removed."""
-        prev_rev = set()
-        updated = []
-        for reviewer in previous_reviewers:
-            if "user" in reviewer:
-                prev_rev.add(f"user:{reviewer['user']}")
-            elif "group" in reviewer:
-                prev_rev.add(f"group:{reviewer['group']}")
-        for reviewer in new_reviewers:
-            if "user" in reviewer:
-                if f"user:{reviewer['user']}" not in prev_rev:
-                    updated.append(reviewer)
-            elif "group" in reviewer:
-                if f"group:{reviewer['group']}" not in prev_rev:
-                    updated.append(reviewer)
+        """Determine reviewers change type: added, removed, updated, unchanged."""
 
-        # NOTE this just supports adding OR removing at a time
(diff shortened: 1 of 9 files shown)


invenio_requests/services/requests/components.py
```diff
@@ -58,57 +58,76 @@ class RequestReviewersComponent(ServiceComponent):
...
+                if (
+                    ("user" in r and f"user:{r['user']}" in removed)
+                    or ("group" in r and f"group:{r['group']}" in removed)
+                )
+            ]
+        elif added and removed:
+            return "updated", list(new_reviewers)
+        else:
+            return "unchanged", list(new_reviewers)
+
```
yashlamba, line 101: nit: this is less readable to me

invenio_requests/services/requests/components.py
```diff
@@ -58,57 +58,76 @@ class RequestReviewersComponent(ServiceComponent):
...
+    def _validate_reviewers(self, new_reviewers, saved_reviewers):
         """Validate the reviewers data."""
         reviewers_enabled = current_app.config["REQUESTS_REVIEWERS_ENABLED"]
         reviewers_groups_enabled = current_app.config["USERS_RESOURCES_GROUPS_ENABLED"]
         max_reviewers = current_app.config["REQUESTS_REVIEWERS_MAX_NUMBER"]
 
-        if not reviewers_enabled:
+        if (new_reviewers or saved_reviewers) and not reviewers_enabled:
```
yashlamba, line 108: question: why are we doing this?

If you add a reviewer and disable the feature for any reason, any request update will fail. IMO, update should check only what is being updated, nothing existing.

invenio_requests/services/requests/components.py
```diff
@@ -58,57 +58,76 @@ class RequestReviewersComponent(ServiceComponent):
...
-        if len(reviewers) > max_reviewers:
+        if len(new_reviewers) > max_reviewers:
             raise ValidationError(
                 _(f"You can only add up to {max_reviewers} reviewers.")
             )
 
-    def update(self, identity, data=None, record=None, **kwargs):
+    def update(self, identity, data=None, record=None, uow=None, **kwargs):
         """Update the reviewers of a request."""
-        if reviewers := data.get("reviewers", None):
```
yashlamba, line 106: ref: this needed to be for the fix
```python
if "reviewers" in data:
    reviewers = data.get("reviewers", [])
```

zzacharo:
> Small suggestion, rest is good.

I kept the old implementation of `ensure_no_duplicates()` because we need to maintain the original order of items

PR #503, opened 2025-10-21
feat(events): deep links

### LINKED ISSUE #494: Enable sharing of comments via a deep-link
## Context

At CERN, we have communities that were creating persistent URLs to comments. In that paradigm, a comment is more of a peer review, which after submission, is immutable to changes.

The feature aims to add a way to "copy the link" to a comment, similar to GitHub, so that this can be shared with others or referenced in comments.

## Tasks

- [x] Frontend design of sharing a comment.
- [x] Frontend design of jumping to a comment from a link.
- [x] Implementation

palkerecsenyi:
Closes #494 

---

* Added a "Copy link" button to events/comments in the UI, which copies the current URL with a hash containing the event ID.
* When TimelineFeed is mounted, it sees if there's a relevant hash in the URL, and requests the server to return the page that contains that event (instead of just requesting the first page).
* Added an argument to the request event `search` route to return the page containing a given event instead of a specific numbered page. The actual page number chosen is returned by the server (see https://github.com/inveniosoftware/invenio-records-resources/pull/656). This is calculated by counting the number of records older than the specified one and dividing by the page size.

## Screenshots

image

This copies the following link:

```
https://127.0.0.1:5000/communities/pal/requests/8642288d-5938-4eb7-a435-728d7266aa09#commentevent-03a3b14c-1f52-46fd-a60a-5b00ceefd7a6
```

And shows up like this when navigated to:

image

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
@@ -11,6 +11,7 @@ export const SUCCESS = "timeline/SUCCESS";
 export const HAS_ERROR = "timeline/HAS_ERROR";
 export const IS_REFRESHING = "timeline/REFRESHING";
 export const CHANGE_PAGE = "timeline/CHANGE_PAGE";
+export const MISSING_REQUESTED_EVENT = "timeline/MISSING_REQUESTED_EVENT";
 
 class intervalManager {
   static IntervalId = undefined;
@@ -25,26 +26,28 @@ class intervalManager {
   }
 }
 
-export const fetchTimeline = (loadingState = true) => {
+export const fetchTimeline = (focusEventId = undefined) => {
   return async (dispatch, getState, config) => {
     const state = getState();
     const { size, page, data: timelineData } = state.timeline;
 
-    if (loadingState) {
-      dispatch({
-        type: IS_LOADING,
-      });
-    }
     dispatch({
       type: IS_REFRESHING,
     });
 
     try {
-      const response = await config.requestsApi.getTimeline({
-        size: size,
-        page: page,
-        sort: "oldest",
-      });
+      let response;
+      if (focusEventId) {
+        response = await config.requestsApi.getTimelineFocused(focusEventId, {
+          size: size,
+        });
+      } else {
+        response = await config.requestsApi.getTimeline({
+          size: size,
+          page: page,
+          sort: "oldest",
+        });
+      }
 
       // Check if timeline has more events than the current state
       const hasMoreEvents = response.data?.hits?.total > timelineData?.hits?.total;
@@ -60,6 +63,23 @@ export const fetchTimeline = (loadingState = true) => {
           const response = await config.requestsApi.getRequest();
           dispatch(updateRequest(response.data));
         }
+      }
+
+      if (response.data.page !== page) {
+        // If a different page was returned (e.g. a specific event ID was requested) we need to update it.
+        // This will _not_ trigger a reload of the timeline.
+        dispatch({
+          type: CHANGE_PAGE,
+          payload: response.data.page,
+        });
+      }
+
+      if (focusEventId && !response.data.hits.hits.some((h) => h.id === focusEventId)) {
+        // Show a warning if the event ID in the hash was not found in the response list of events.
+        // This happens if the server cannot find the requested event.
+        dispatch({
+          type: MISSING_REQUESTED_EVENT,
+        });
       }
 
       dispatch({
@@ -81,6 +101,9 @@ export const setPage = (page) => {
       type: CHANGE_PAGE,
       payload: page,
     });
+    dispatch({
+      type: IS_LOADING,
+    });
 
     await dispatch(fetchTimeline());
   };
@@ -99,12 +122,15 @@ const timelineReload = (dispatch, getState, config) => {
 
   if (concurrentRequests) return;
 
-  dispatch(fetchTimeline(false));
+  dispatch(fetchTimeline());
 };
 
-export const getTimelineWithRefresh = () => {
+export const getTimelineWithRefresh = (focusEventId) => {
   return async (dispatch, getState, config) => {
-    dispatch(fetchTimeline(true));
+    dispatch({
+      type: IS_LOADING,
+    });
+    dispatch(fetchTimeline(focusEventId));
     dispatch(setTimelineInterval());
   };
 };
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvents/TimelineCommentEvent.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvents/TimelineCommentEvent.js
@@ -6,7 +6,7 @@
 
 import { i18next } from "@translations/invenio_requests/i18next";
 import PropTypes from "prop-types";
-import React, { Component } from "react";
+import React, { Component, createRef } from "react";
 import { Image } from "react-invenio-forms";
 import Overridable from "react-overridable";
 import { Container, Dropdown, Feed, Icon } from "semantic-ui-react";
@@ -16,6 +16,8 @@ import { RichEditor } from "react-invenio-forms";
 import RequestsFeed from "../components/RequestsFeed";
 import { TimelineEventBody } from "../components/TimelineEventBody";
 import { toRelativeTime } from "react-invenio-forms";
+import { isEventSelected } from "./utils";
+import { RequestEventsLinksExtractor } from "../api/InvenioRequestEventsApi.js";
 
 class TimelineCommentEvent extends Component {
   constructor(props) {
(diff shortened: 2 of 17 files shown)


invenio_requests/services/events/service.py
```diff
@@ -236,38 +237,59 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
             search_preference,
             permission_action="unused",
-            extra_filter=dsl.Q("term", request_id=str(request.id)),
+            extra_filter=request_filter,
+            versioning=False,
             **kwargs,
         )
-        search_result = search.execute()
 
+        if include_event_id is not None:
```
zzacharo: Can we add the `include_event_id` as new Param to isolate the behaviour of the filtering?

invenio_requests/services/events/service.py
```diff
@@ -236,38 +237,59 @@ def delete(self, identity, id_, revision_id=None, uow=None):
         return True
 
     def search(
-        self, identity, request_id, params=None, search_preference=None, **kwargs
+        self,
+        identity,
+        request_id,
+        include_event_id=None,
```
zzacharo: E.g. below maybe define a new param instead?

ntarocco: Looking at the code, I would go for a new specific API method. My feeling is that this is not really a search: it should allow extra search params (or should it?).
If I fetching one specific comment, filtering or sorting should be ignored.
WDYT?

```python
def jump_to(self, identity, request_id, event_id):
    ....
```

invenio_requests/services/events/service.py
```diff
@@ -236,29 +238,59 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
+                num_older_than_event = search.filter(
+                    "range", created={"lt": event.created}
+                ).count()
+                page = num_older_than_event // params["size"] + 1
+                # Re run the pagination param interpreter to update the search with the new page number
+                params.update(page=page)
+                search = PaginationParam(self.config.search).apply(
+                    identity, search, params
+                )
+            except sqlalchemy.exc.NoResultFound:
```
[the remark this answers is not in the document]
ntarocco: Totally agree, at the same time, it would be very weird that we don't signal it in any way: the user will consider that there is a bug in jumping to a specific comment.
Can we render the comments page, and at the same time show a message that the comment was not found?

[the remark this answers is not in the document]
palkerecsenyi: Sure that sounds good, I can implement a message!

palkerecsenyi:
Ahh okay, I am happy to try this and we can see what it looks like! My main concern is making sure we keep to minimal code duplication, and this would probably do that.

invenio_requests/services/events/service.py
```diff
@@ -241,25 +243,87 @@ def search(
...
         # Prepare and execute the search
-        search = self._search(
+        search_result = self._search(
             "search",
             identity,
             params,
             search_preference,
             permission_action="unused",
             extra_filter=dsl.Q("term", request_id=str(request.id)),
+            versioning=False,
```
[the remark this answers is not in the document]
palkerecsenyi: thanks!

invenio_requests/services/events/service.py
```diff
@@ -241,25 +243,86 @@ def search(
...
         )
-        search_result = search.execute()
 
+    def focussed_list(
+        self,
+        identity,
+        request_id,
+        focus_event_id,
+        page_size,
+        expand=False,
```
zzacharo:
```suggestion
        expand=expand,
```
We should still allow the expand parameter as this controls the expansion of entities like `created_by`

invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/state/actions.js
```diff
@@ -99,12 +123,15 @@ const timelineReload = (dispatch, getState, config) => {
...
 };
 
-export const getTimelineWithRefresh = () => {
+export const getTimelineWithRefresh = (focusEventId) => {
   return async (dispatch, getState, config) => {
-    dispatch(fetchTimeline(true));
+    dispatch({
+      type: IS_LOADING,
+    });
+    dispatch(fetchTimeline(focusEventId, true));
```
zzacharo: you forgot here?
```suggestion
    dispatch(fetchTimeline(focusEventId));
```

invenio_requests/services/events/config.py
```diff
@@ -67,9 +72,11 @@ class RequestEventLink(Link):
     """Link variables setter for RequestEvent links."""
 
     @staticmethod
-    def vars(record, vars):
+    def vars(obj, vars):
         """Variables for the URI template."""
-        vars.update({"id": record.id, "request_id": record.request_id})
+        request_type = current_request_type_registry.lookup(vars["request_type"])
```
zzacharo: 🚀

invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvents/TimelineCommentEvent.js
```diff
@@ -75,32 +102,37 @@ class TimelineCommentEvent extends Component {
     }
     return (
       <Overridable id={`TimelineEvent.layout.${this.eventToType(event)}`} event={event}>
-        <RequestsFeed.Item>
+        <RequestsFeed.Item
+          id={`commentevent-${event.id}`}
```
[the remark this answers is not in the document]
palkerecsenyi: I've resolved this as much as possible but I think we still have to hardcode the `commentevent` for the [function](https://github.com/inveniosoftware/invenio-requests/pull/503/files#diff-ee21c616072d63d01d5813b4a3f5b03a52ffb852357b4f1643c49812aac49de8R15-R23) that retrieves the event ID from the URL on page load.

invenio_requests/assets/semantic-ui/js/invenio_requests/timelineEvents/utils.js
```diff
@@ -0,0 +1,23 @@
...
+  const eventUrl = new URL(new RequestEventsLinksExtractor(event.links).eventHtmlUrl);
+  const currentUrl = new URL(window.location.href);
+  return eventUrl.hash === currentUrl.hash;
+};
+
+export const getEventIdFromUrl = () => {
+  const currentUrl = new URL(window.location.href);
+  const hash = currentUrl.hash;
+  let eventId = null;
+  if (hash.startsWith("#commentevent-")) {
```
zzacharo:
```suggestion
  const commentPrefix = "#commentevent-"
  if (hash.startsWith(commentPrefix)) {
    eventId = hash.substring(commentPrefix.length);
```

invenio_requests/services/events/service.py
```diff
@@ -45,14 +45,14 @@ def expandable_fields(self):
         """Get expandable fields."""
         return [EntityResolverExpandableField("created_by")]
 
-    def links_item_tpl_for_request(self, request):
+    def links_item_tpl_for_request_type(self, request_type: str):
```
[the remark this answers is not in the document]
zzacharo: @ntarocco @slint @kpsherva any objection to this?

invenio_requests/notifications/generators.py
```diff
@@ -60,6 +60,11 @@ def __call__(self, notification, recipients: dict):
             identity=system_identity,
             extra_filter=dsl.Q("term", request_id=request["id"]),
         )
+        # Override the _links_item_tpl as done by all other RecordList-returning methods in the
+        # RequestEvents service.
+        request_events._links_item_tpl = (
```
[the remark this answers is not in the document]
zzacharo: @fenekku any idea to make this more elegant maybe?

invenio_requests/services/events/config.py
```diff
@@ -90,6 +96,7 @@ class RequestEventsServiceConfig(RecordServiceConfig, ConfiguratorMixin):
     # ResultItem configurations
     links_item = {
         "self": RequestEventLink("{+api}/requests/{request_id}/comments/{id}"),
+        "self_html": RequestEventLink("{+ui}/requests/{request_id}#commentevent-{id}"),
```
fenekku: Note to self: going to have to implement anchor to fully move away from Link
 
https://github.com/inveniosoftware/invenio-base/blob/master/invenio_base/urls/helpers.py#L22

fenekku (approved):
Thanks for that and kudos for going through that code. I was trying to work out "nicer" ways of dealing with all this all morning and nothing was particularly easier (more a commentary on the codebase we have 😄 ).

invenio_requests/services/events/service.py
```diff
@@ -241,24 +253,118 @@ def search(
...
+            permission_action="unused",
+            extra_filter=dsl.Q("term", request_id=str(request.id)),
+            versioning=False,
+        )
+
+        page = 1
+        if focus_event is not None:
+            self.require_permission(
+                identity, "read", request=request, event=focus_event
+            )
```
slint: minor: if we check above that the `focus_event` belongs to this request, we don't need to have another permission check here.

invenio_requests/notifications/generators.py
```diff
@@ -57,6 +57,7 @@ def __call__(self, notification, recipients: dict):
 
         # fetching all request events to get involved users
         request_events = current_events_service.scan(
+            request_id=request["id"],
```
slint: shelve: looking at this now, it's not great that we go through **all** of the comments to find the participants (imagine 10k comments between just 2 participants)... we should instead:

- Keep this, but only fetch the `created_by.user` field via the `source` search parameter
- Do an aggregation query (`terms` bucket) on `created_by.user` to fetch the unique user IDs

invenio_requests/services/events/service.py
```diff
@@ -117,7 +130,7 @@ def create(
             identity,
             event,
             schema=schema,
-            links_tpl=self.links_item_tpl,
+            links_tpl=self.links_item_tpl_for_request_type(str(request.type)),
```
zzacharo: @palkerecsenyi is this still needed after overridding the `scan` method?

invenio_requests/services/events/service.py
```diff
@@ -43,6 +46,16 @@ def expandable_fields(self):
         """Get expandable fields."""
         return [EntityResolverExpandableField("created_by")]
 
+    def links_item_tpl_for_request_type(self, request_type: str):
```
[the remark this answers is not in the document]
palkerecsenyi: I've now reimplemented it as `links_tpl_factory`!

Issue #490, opened 2025-10-14
Enable single-threaded discussion per comment


zzacharo:
## Context 

This feature aims to add a one level thread level per comment. The incentive behind this is that a comment is considered a thorough review of a publication. In that context, having a conversation between the authors and the reviewer regarding the submitted review makes sense. The one-level single thread is a compromise between "no replies per comment" and "nested threading" that still keeps the focus of the commenting on the submitted resource but facilitates the interaction between authors and reviewers.

## Existing design ideas

- BNL prototype: https://github.com/inveniosoftware/product-rdm/discussions/232#discussioncomment-13865342
- Generated mockup
<img width="484" height="427" alt="Image" src="https://github.com/user-attachments/assets/aa3eee85-d3b0-473b-bda5-3501efcc1f8f" />

- GH discussion style e.g. https://github.com/inveniosoftware/product-rdm/discussions/232#discussioncomment-14224162

## Tasks

- [ ] RFC design
  - [ ] Backend support of single-level comments per comment
  - [ ] Frontend design of comment replies
- [ ] Implementation

palkerecsenyi:
We are currently considering two designs, see the mockups below:

https://share.balsamiq.com/c/eA8FM6u5d2FYnHVWN9TNtf.jpg

PR #505, opened 2025-10-27
feat(events): support request event children

zzacharo:
* closes https://github.com/inveniosoftware/invenio-requests/issues/490


invenio_requests/services/events/config.py
```diff
@@ -35,6 +35,59 @@ def id(self):
...
+                    context=dict(
+                        identity=self._identity,
+                        record=record,
+                    ),
+                )
+                if self._links_item_tpl:
+                    projection["links"] = self._links_item_tpl.expand(
+                        self._identity, record
+                    )
+                child.update(projection)
```
zzacharo: We need to override this to apply record serialization for `children_preview` array...

PR #507, opened 2025-11-04
tests: set transition range for python versions (rdm)

fenekku:


fenekku:
main workflow does it, closing

PR #514, opened 2025-11-19
feat(editor): add draft save status message

### LINKED ISSUE #493: Autosave the current draft comment written in the textarea
## Context

Users tend to add extensive comments directly via the web application, and often a wrong misclick can lead to the loss of a significant amount of work.

## Tasks

- [ ] Frontend design of saving/restoring a comment from `localStorage` to avoid losing work.
- [ ] Implementation

## Resources

- Check [Autosave](https://www.tiny.cloud/docs/tinymce/latest/autosave/) plugin

palkerecsenyi:
Closes #493

---

* Added a message to show the status of the draft saving. To give a more natural feel, it shows up 500ms after the last input from the user (but the draft is actually saved immediately).
* See https://github.com/inveniosoftware/invenio-app-rdm/pull/3245 for CSS changes

https://github.com/user-attachments/assets/f9b2c734-53c1-437e-aa61-549d056bcf6d

On mobile:

image

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/components/TimelineDraftSavingStatus.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/components/TimelineDraftSavingStatus.js
@@ -1,0 +1,43 @@
+// This file is part of InvenioRequests
+// Copyright (C) 2025 CERN.
+//
+// Invenio Requests is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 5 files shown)


palkerecsenyi:
We have decided that this message gives too much confidence to users given the fragility of Local Storage and the fact that changes are only saved to the browser. So for now we won't implement this.

Issue #498, opened 2025-10-14
Apply a grace period for comment actions (Lock discussion - comment actions)


zzacharo:
## Context

Usually after adding a comment/review, the receivers of the comment start working on the posted review. There should be a configurable grace period after which specific actions on the comment and in the request discussion should be disabled.

## Tasks

- [ ] Frontend design of comment actions disable (who can disable, how)
   - [ ] Check mockups in https://balsamiq.cloud/ssg4nq2/p68p1b0/r4004 
- [ ]  Backend design of "locking a conversation". Ensure proper permissions in place to fine tune comment actions.
     - [ ] Curators of a community will be able to lock an entire conversation.
     - [ ] We need to decide on the request level who can lock the conversation? Receiver always? 
     - [ ] The default implementation will lock any action to the comments (create/update/delete etc.). In threaded comments, it might be relevant to allow replies to comments even after locking. This should impact only the addition of a specific permission when adding a reply to a comment e.g `can_reply = can_create` and each instance will override accordingly.
     - [ ] A log event should indicate that a lock/unlock of the conversation has happened, from whom and when.
- [ ] Implementation

PR #518, opened 2025-11-24
feat(requests): Add conversation locking functionality

sakshamarora1:
closes: #498

### DIFF
--- a/invenio_requests/services/requests/service.py
+++ b/invenio_requests/services/requests/service.py
@@ -10,7 +10,6 @@
 
 """Requests service."""
 
-from flask import current_app, flash
 from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services import RecordService, ServiceSchemaWrapper
 from invenio_records_resources.services.base import LinksTemplate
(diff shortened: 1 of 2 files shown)


invenio_requests/services/events/service.py
```diff
@@ -72,7 +73,19 @@ def create(
...
+        try:
+            self.require_permission(
+                identity, "create_comment", request=request, event_type=event_type
+            )
+        except PermissionDeniedError:
+            if request.get("is_locked", False):
+                raise RequestLockedError(
+                    description=_("Commenting is now locked for this conversation.")
+                )
+            else:
```
zzacharo: any particular case that this condition can be true? Messed up permissions?

sakshamarora1: Yes, and another thing is that we can show a nicer message for API calls.

invenio_requests/services/requests/service.py
```diff
@@ -313,3 +315,31 @@ def search_user_requests(
...
         )
+
+    @unit_of_work()
+    def lock_request(self, identity, id_, uow=None):
+        """Lock a request."""
+        request = self.read(identity, id_)
+        self.require_permission(identity, "lock_request", request=request._record)
+
+        if request.data["is_locked"]:
+            raise RequestLockedError(description="The request is already locked.")
```
zzacharo: translations?

PR #520, opened 2025-11-25
feat(comment-replies): add single threading on comments

### LINKED ISSUE #490: Enable single-threaded discussion per comment

zzacharo:
closes #490 
addresses https://github.com/inveniosoftware/rfcs/issues/107


invenio_requests/records/validators.py
```diff
@@ -0,0 +1,28 @@
...
+    """Validate that an event type allows children before setting parent_id.
+
+    :param event: The RequestEvent instance to validate.
+    :raises ChildrenNotSupportedError: If the event type doesn't support children
+                                       but has a parent_id set.
+    """
+    if event.parent_id is not None:
+        # Check if event type supports children
+        if hasattr(event, "type") and event.type:
+            if not getattr(event.type, "allow_children", False):
```
palkerecsenyi: We could refactor this to use guard clauses instead for slightly better readability, just to avoid nesting the if statements. But this is also not too big of an issue

invenio_requests/services/events/config.py
```diff
@@ -75,6 +109,8 @@ def vars(obj, vars):
         """Variables for the URI template."""
         request_type = current_request_type_registry.lookup(vars["request_type"])
         vars.update({"id": obj.id, "request_id": obj.request_id})
+        parent_id = obj.parent_id if obj.parent_id else obj.id
+        vars.update({"parent_id": parent_id})
         vars.update(request_type._update_link_config(**vars))
```
palkerecsenyi:
```suggestion
        request_type = current_request_type_registry.lookup(vars["request_type"])
        parent_id = obj.parent_id if obj.parent_id else obj.id
        vars.update(
            {"id": obj.id, "request_id": obj.request_id, "parent_id": parent_id}
        )
        vars.update(request_type._update_link_config(**vars))
```

invenio_requests/services/events/config.py
```diff
@@ -39,6 +40,28 @@ def id(self):
...
+            children = self._data.get("children", [])
+            if children:
+                # Batch resolve all children
+                self._fields_resolver.resolve(self._identity, children)
+                # Expand each child
+                for child in children:
+                    fields = self._fields_resolver.expand(self._identity, child)
+                    child["expanded"] = fields
+
+        return self._data
```
slint: This is probably not needed anymore, since the only cases where we have `children` are on search-related results (i.e. `.get_comment_replies` and `.timeline`).

invenio_requests/services/events/service.py
```diff
@@ -154,6 +175,37 @@ def create(
             expand=expand,
         )
 
+    @unit_of_work()
+    def reply(
```
slint: minor: I think we can delete this, and just use `create` in the resource call since we also do the permission check there.

invenio_requests/services/events/service.py
```diff
@@ -120,7 +139,7 @@ def create(
             uow=uow,
         )
 
-        # Persist record (DB and index)
+        # Persist record (DB and index) - top-level events only
```
slint:
```suggestion
        # Persist record (DB and index)
```

Not the case anymore

invenio_requests/errors.py
```diff
@@ -90,3 +90,36 @@ def description(self):
     def __str__(self):
         """Return str(self)."""
         return self.description
+
+
+class ChildrenNotSupportedError(Exception):
```
slint: check if this needs to be handled by the resource `error_handlers`

invenio_requests/errors.py
```diff
@@ -90,3 +90,36 @@ def description(self):
...
+
+    def __init__(self, message=None):
+        """Constructor.
+
+        :param message: Optional custom error message.
+        """
+        self.message = message or (
+            "Nested children are not allowed. "
+            "You cannot reply to a comment that is already a reply. "
+            "Only one level of parent-child relationships is supported."
```
slint: this needs i18n

PR #487, opened 2025-09-30
context plus utc

utnapischtim:
- refactor: move to context_schema
- fix(chore): DeprecationWarning stdlib

PR #522, opened 2025-12-01
deprecate link

fenekku:
- :warning: depends on [invenio-records-resources PR](https://github.com/inveniosoftware/invenio-records-resources/pull/659) for anchor
- :warning: will require a major bump because we are removing old constructs to generate links used in invenio-rdm-records and invenio-app-rdm by new ones that will need additions in those packages.
- **refactor!: replace deprecated Link usage [+]**
- **chore(tests): replace deprecated es_clear fixture by search_clear**

### DIFF
--- a/invenio_requests/services/links.py
+++ b/invenio_requests/services/links.py
@@ -70,7 +70,7 @@ class RequestTypeDependentEndpointLink(EndpointLink):
         "request", "request_type", "request_event" that an EndpointLink
         defined on a RequestType can rely on.
         """
-        ctx = copy.deepcopy(context)
+        ctx = context.copy()  # deepcopy can cause problems here
         ctx["request"] = self._request_retriever(obj, ctx)
         ctx["request_type"] = self._request_type_retriever(obj, ctx)
         ctx["request_event"] = self._request_event_retriever(obj, ctx)
@@ -167,10 +167,10 @@ class ActionsEndpointLinks:
         links = {}
         request = obj
 
+        ctx = context.copy()
         for action in request.type.available_actions:
             if action in [request.type.create_action, request.type.delete_action]:
                 continue
-            ctx = context.copy()
             ctx["action"] = action
             if self._endpoint_link.should_render(request, ctx):
                 links[action] = self._endpoint_link.expand(request, ctx)
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,14 +26,14 @@ packages = find:
 python_requires = >=3.7
 zip_safe = False
 install_requires =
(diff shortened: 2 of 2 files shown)


invenio_requests/services/requests/config.py
```diff
@@ -103,16 +111,28 @@ class RequestsServiceConfig(RecordServiceConfig, ConfiguratorMixin):
...
+        "self": RequestEndpointLink("requests.read"),
+        "self_html": RequestTypeEndpointLink(
+            "self_html",
+            params=["request_pid_value"],
+            vars=lambda obj, vars: vars.update({"request_pid_value": obj.id}),
+        ),
+        "comments": RequestEventsEndpointLink("request_events.create"),
+        "timeline": RequestEventsEndpointLink("request_events.search"),
+        "timeline_focused": RequestEventsEndpointLink("request_events.focused_list"),
+        "actions": ActionsEndpointLinks(
```
fenekku: This is a pattern I will be bringing to other places if works well: actions are part of the links_item and we can fake the interface to make them produce another layer of dict of links. It removes the need for RequestLinksTemplate (which shrinks the code surface for link generation which makes it easier to understand and work with).

invenio_requests/services/events/config.py
```diff
@@ -232,19 +218,49 @@ class RequestEventsServiceConfig(RecordServiceConfig, ConfiguratorMixin):
...
+            ),
+        ),
+        "replies": RequestCommentEndpointLink(
+            "request_events.get_replies",
+            # this runs after the RequestCommentEndpointLink.vars so it
+            # takes precedence
+            vars=lambda obj, vars: vars.update(
+                # This translates previous logic, but does it make sense
+                # to list the replies to the parent if it exists and not
+                # always to the current comment?
```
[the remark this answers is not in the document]
zzacharo: The point of having for both endpoints only the `parent_id` is that any link related to a comment reply makes sense only if the comment is a parent. I intentionally always generated the link even for child comments pointing to the parent so that the consumer of the REST API do not have to check for existtence in order to reply. If we change the behaviour @palkerecsenyi do we have to change the frontend code?

invenio_requests/services/events/service.py
```diff
@@ -169,7 +169,7 @@ def create(
             event,
             schema=schema,
             links_tpl=self.links_tpl_factory(
-                self.config.links_item, request_type=str(request.type)
+                self.config.links_item, request=request, request_type=request.type
```
fenekku: We now also provide the request, so that endpoints defined on the RequestType can make use of that information.

utnapischtim:
combined with #487 to #560 , so closed

PR #539, opened 2025-12-17
fix(replies-permissions): pass correct reply permissions

zzacharo:


invenio_requests/assets/semantic-ui/js/invenio_requests/timeline/TimelineFeed.js
```diff
@@ -201,7 +201,10 @@ class TimelineFeed extends Component {
                 </RequestsFeed>
               )}
 
-              <TimelineCommentEditor userAvatar={userAvatar} />
+              <TimelineCommentEditor
+                userAvatar={userAvatar}
+                canCreateComment={permissions.can_create_comment}
```
palkerecsenyi: I have also fixed this in https://github.com/inveniosoftware/invenio-requests/pull/538

palkerecsenyi (approved):

PR #546, opened 2026-01-16
fix(timeline): support showing replies preview for timeline_focused endpoint

palkerecsenyi:


invenio_requests/services/events/service.py
```diff
@@ -424,12 +415,17 @@ def focused_list(
             pass
 
         params = {"sort": "oldest", "size": page_size}
+
+        # TODO: this needs to be adpated to focus on links to child comments
```
zzacharo: hanging comment?

PR #548, opened 2026-01-21
Deep link into replies

### LINKED ISSUE #542: Deep links for replies
In #494 we created a feature for copying a deep link to a comment. We need to also enable this for replies, which will possibly also involve creating another backend route (we needed `timeline_focused` for the comments)

OliverGeneser:
Closes #542 

### Description

> This now makes it possible to deep link into replies


sakshamarora1 (commented):
I've added some initial doubts I have. But it would be better to have a screen recording or screenshot of how it looks like to get more context.

PR #606, opened 2026-05-04
fix(comments): ensure "show less" button is hidden for comments <200px height

### LINKED ISSUE #605: ui: "Show less" button shown on non-collapsible comments when `expandedByDefault` is set
When setting the [`expandedByDefault`](https://github.com/inveniosoftware/invenio-requests/blob/adc4bacb3dbfbccc79f0dfbffde67430177f8e7e/invenio_requests/assets/semantic-ui/js/invenio_requests/components/TimelineEventBody.js#L119) prop to `true` on `TimelineEventBody`(`Container`) via React Overridable / `parametrize`, the "Show less" button is shown on comments that are too short to need collapsibility. When the button is clicked, it disappears:

https://github.com/user-attachments/assets/38a8d612-13f7-4fb3-9a00-a40a445ac87b

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-requests/issues/605

---

**Before**

https://github.com/user-attachments/assets/20a07ce3-2951-4e1f-b7eb-b3491c949a70

**After**

https://github.com/user-attachments/assets/99f28ddf-6b98-4aba-9ff0-1138ad59ef6c

### DIFF
--- a/invenio_requests/assets/semantic-ui/js/invenio_requests/components/TimelineEventBody.js
+++ b/invenio_requests/assets/semantic-ui/js/invenio_requests/components/TimelineEventBody.js
@@ -158,8 +158,14 @@ const TimelineEventBodyContainer = ({
   useEffect(() => {
     if (!collapsible) return;
 
+    const el = refInner.current;
     if (expanded) {
-      setIsOverflowing(true);
+      // The collapsed comment height is 200px. If the comment is too short to need expanding but `expanded` is true
+      // nonetheless, `el.scrollHeight` will still be less than 200, so the comment will not be overflowing.
+      // This ensures the "Show less" button is hidden for comments where `expandedByDefault` is true but the comment
+      // is too short to need expanding/collapsing.
+      setIsOverflowing(el.scrollHeight > 200);
+      // To prevent race conditions, we disable the resize observer once the comment is expanded.
       return;
     }
 
@@ -169,7 +175,6 @@ const TimelineEventBodyContainer = ({
       setIsOverflowing(el.scrollHeight > el.clientHeight);
     });
 
-    const el = refInner.current;
     if (!el) return;
     resizeObserver.observe(el);
     return () => {
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@
 Changes
 =======
 
+Version v12.6.1 (released 2026-05-05)
+
+- fix(comments): ensure "show less" button is hidden for comments <200px height
+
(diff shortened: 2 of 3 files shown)


kpsherva (approved):

sakshamarora1 (approved):

OliverGeneser (approved):

invenio_requests/assets/semantic-ui/js/invenio_requests/components/TimelineEventBody.js
```diff
@@ -158,8 +158,14 @@ const TimelineEventBodyContainer = ({
...
     if (!collapsible) return;
 
+    const el = refInner.current;
     if (expanded) {
-      setIsOverflowing(true);
+      // The collapsed comment height is 200px. If the comment is too short to need expanding but `expanded` is true
+      // nonetheless, `el.scrollHeight` will still be less than 200, so the comment will not be overflowing.
+      // This ensures the "Show less" button is hidden for comments where `expandedByDefault` is true but the comment
+      // is too short to need expanding/collapsing.
+      setIsOverflowing(el.scrollHeight > 200);
```
zzacharo, line 167: isnt this configurable? the comment height?

palkerecsenyi: Yes, it is configurable via a LESS variable [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/96f9d528742b172194899e09917a113d26ef0e0d/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/feed.variables#L2). I guess there is no easy way to get this value in the frontend JS, so maybe we could use an inline style instead of the LESS and define the height via a prop (so that react-overridable can be used to override it). But then originally we [decided](https://github.com/inveniosoftware/invenio-requests/pull/543#discussion_r2731753291) we wanted to keep it as a LESS variable.

We are using the `el.scrollHeight > el.clientHeight` trick inside the `ResizeObserver` a few lines down, but of course this wouldn't work here since the container is unbounded so the condition would always be true. We don't have a fixed reference frame to compare the content height to, which is why we need the actual pixel value. At least I can't think of a clever way to do it without.

zzacharo: Alright, lets start like that then and we can reevaluate in the future if needed

zzacharo (approved):

PR #607, opened 2026-05-21
fix(reviewers): have `payload.event` reflect the actual event type

max-moser:
Previously, `ReviewersUpdatedType` events were always created with `payload.event = "reviewers_updated"`, regardless of the actual event type (added/removed/updated/unchanged).
In this PR, we use the determined event type instead to add more context to the event.

### DIFF
--- a/invenio_requests/services/requests/components.py
+++ b/invenio_requests/services/requests/components.py
@@ -157,7 +157,7 @@ class RequestReviewersComponent(ServiceComponent):
             if not event_type == "unchanged":
                 event = ReviewersUpdatedType(
                     payload=dict(
-                        event="reviewers_updated",
+                        event=f"reviewers_{event_type}",
                         content=_(f"{event_type} a reviewer"),
                         reviewers=updated_reviewers,
                     )


invenio_requests/services/requests/components.py
```diff
@@ -157,7 +157,7 @@ def update(self, identity, data=None, record=None, uow=None, **kwargs):
             if not event_type == "unchanged":
                 event = ReviewersUpdatedType(
                     payload=dict(
-                        event="reviewers_updated",
+                        event=f"reviewers_{event_type}",
                         content=_(f"{event_type} a reviewer"),
                         reviewers=updated_reviewers,
```
zzacharo, line 162: I think the problem and why we went for always have `updated` to the event is that we didnt not track what was removed or added. If we do that, then the reviewers object will not depict e.g the removed reviewers.

How do you use this change in your instance @max-moser ?

max-moser: I've recently added notification emails when reviewers are added/removed from a request to `Invenio-Config-TUW`, as that was requested by our data stewards.
For now, I'm just checking the `content`, but that feels very much like a hack; see here: https://gitlab.tuwien.ac.at/crdm/invenio-config-tuw/-/commit/0c8e03edf02afa9496107945ffd5bdea2e82ee63#line_81daeea27_A374

PR #614, opened 2026-06-16
feat(requests): add role grant to shared with me param

palkerecsenyi:
This ensures that membership of a group will also allow requests to be shown in the "Shared with me" tab

### DIFF
--- a/invenio_requests/services/requests/params.py
+++ b/invenio_requests/services/requests/params.py
@@ -108,6 +108,11 @@ class SharedOrMyRequestsParam(ParamInterpreter):
             for need in identity.provides
             if need.method in allowed_need_methods
         ]
+        receiver_grants = [
+            EntityGrant("receiver", need).token
+            for need in identity.provides
+            if need.method in allowed_need_methods
+        ]
         reviewer_grants = [
             EntityGrant("reviewers", need).token
             for need in identity.provides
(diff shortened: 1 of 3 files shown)


invenio_requests/services/requests/params.py
```diff
@@ -108,6 +113,11 @@ def _generate_shared_with_me_query(self, identity):
             for need in identity.provides
             if need.method in allowed_need_methods
         ]
+        role_grants = [
```
zzacharo:
```suggestion
        receiver_grants = [
```

zzacharo:
@max-moser maybe you want to have a look into this as it might simplify your curations module

max-moser:
Basically, that would enable the `rdm-curation` requests showing up in the "shared with me" tab for those users that have the `reviewer` role, without some of the machinery in `Invenio-Curations`?

zzacharo:
> Basically, that would enable the `rdm-curation` requests showing up in the "shared with me" tab for those users that have the `reviewer` role, without some of the machinery in `Invenio-Curations`?

I think as a reviewer you already have that. This PR will enable requests that are happening against a receiver group e.g tuw-curators, to be shown in the users' shared-with-me tab.

max-moser:
Oh yes, that's what I meant.
That group you called `tuw-curators` is called `reviewers` by default; it's different from the (newer) request reviewers feature from `Invenio-Requests`, which I assume you were referring to?

zzacharo:
> Oh yes, that's what I meant. That group you called `tuw-curators` is called `reviewers` by default; it's different from the (newer) request reviewers feature from `Invenio-Requests`, which I assume you were referring to?

Ah I forgot about it, then yes it will enable you what you said

invenio_requests/services/requests/params.py
zzacharo: @palkerecsenyi can you apply the suggestoin here so I can merge?

